PR #141, opened 2026-08-25
improved wifi connection backend, improved status/settings page

moldbuster:
## Description

I was having difficulties connecting to SomnoTrace because I had multiple AP's broadcasting the same SSID, but SomnoTrace chose a weak one, causing the pages to load really slow with errors.  So I improved the wifi connect strategy and it will now pick the strongest one.  Also if multiple SSID's are set, it will start from the first and if they are not found will continue going down the list.  I've improved the status/settings page to improve the usability.

## Summary

On networks where multiple access points broadcast the **same SSID** (mesh / extenders),
SomnoTrace could associate with an **arbitrary** AP — the supplicant's default fast-scan
grabbed the first matching beacon during its channel sweep, discarding the scan's
signal-strength ranking. This randomly landed the device on a distant BSS (observed at
−94 dBm with closer siblings available), producing TCP stalls that surfaced as 502s through
a fronting reverse proxy and an empty dashboard on single-shot page loads.

This PR pins association to the **strongest scanned AP**, makes multi-network fallback
behaviour explicit (**strict slot priority**: Network #1 first), and adds at-a-glance
signal/connected indicators to the portal UI.

## Firmware (`main/net_provision.c`, `main/net_provision.h`)

- **Pin association to the scanned BSS**: `try_single_ssid()` accepts the candidate
  `wifi_ap_record_t` and sets `bssid_set`/BSSID/channel on attempt 1 (attempts 2–3 fall back
  to un-pinned so a sibling AP can take over if the pinned one died).
- **Explicit authmode threshold**: `WPA2_PSK` when a password is set, `OPEN` otherwise
  (removes the supplicant's auto-raise warning).
- **PMF capable** (not required): future-proofing for WPA2/WPA3 mixed-mode APs.
- **Scan record cap 32 → 60**: dense RF environments could truncate before the strongest
  same-SSID record was ranked.
- **Slot-priority connect order**: cross-slot RSSI sorting removed — Network # 1 is always
  tried first (3 attempts, 5 s apart) before Network # 2; within a slot the strongest AP wins.

## Portal UI (`main/portal.html`)

- Wi-Fi dropdown options now start with a coloured strength circle 🟢/🟡/🔴 (≥ −60 /
  −74…−60 / ≤ −75 dBm).
- Settings blocks show a green **(CONNECTED)** marker on the configured network that matches
  the current association.
- Status → System Status lists each configured network as `Wi-Fi Network #N`; the connected
  row shows its live RSSI with the whole value coloured by signal strength; inactive rows
  are greyed out.

## Checklist

### DIFF
--- a/main/net_provision.c
+++ b/main/net_provision.c
@@ -424,7 +424,10 @@ esp_err_t netprov_init(void)
 /* ------------------------------------------------------------------ */
 /*  STA connect with scan + candidate selection                       */
 /* ------------------------------------------------------------------ */
+/* rec != NULL pins the association to the exact BSS found by the candidate
+ * scan (BSSID + channel).  NULL falls back to supplicant FAST_SCAN by SSID. */
 static esp_err_t try_single_ssid(const char *ssid, const char *pass,
+                                 const wifi_ap_record_t *rec,
                                  char *ip_out, int timeout_ms)
 {
     if (s_portal_mode) return ESP_FAIL;
@@ -435,7 +438,22 @@ static esp_err_t try_single_ssid(const char *ssid, const char *pass,
     wifi_config_t wc = { 0 };
     strlcpy((char *)wc.sta.ssid, ssid, sizeof(wc.sta.ssid));
     strlcpy((char *)wc.sta.password, pass, sizeof(wc.sta.password));
-    wc.sta.threshold.authmode = WIFI_AUTH_OPEN;
+    /* Non-empty password implies at least WPA2 — being explicit is
+     * deterministic and silences the supplicant's auto-raise warning. */
+    wc.sta.threshold.authmode =
+        (pass[0] != '\0') ? WIFI_AUTH_WPA2_PSK : WIFI_AUTH_OPEN;
+    /* PMF capable (not required): keeps WPA2 working, enables WPA2/WPA3
+     * mixed-mode APs. */
+    wc.sta.pmf_cfg.capable = true;
+    if (rec) {
+        memcpy(wc.sta.bssid, rec->bssid, sizeof(wc.sta.bssid));
+        wc.sta.bssid_set = true;
+        wc.sta.channel   = rec->primary;   /* skip the channel sweep */
+        ESP_LOGI(TAG, "pinning '%s' bssid=%02x:%02x:%02x:%02x:%02x:%02x ch=%d rssi=%d",
+                 ssid, rec->bssid[0], rec->bssid[1], rec->bssid[2],
+                 rec->bssid[3], rec->bssid[4], rec->bssid[5],
+                 rec->primary, rec->rssi);
+    }
 
     ESP_ERROR_CHECK(esp_wifi_set_mode(WIFI_MODE_STA));
     ESP_ERROR_CHECK(esp_wifi_set_config(WIFI_IF_STA, &wc));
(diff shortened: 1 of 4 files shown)


moldbuster:
image

An example when I connect to a weak wifi, this shows up

ilyakruchinin:
The example above has no relationship to ESP32/SomnoTrace.
This response was from OpenResty (see text) that the ESP does not use

ilyakruchinin:
`<hr><center>openresty</center> </body>`
This clearly identifies the web server that had the 50X error, and it is not the ESP

moldbuster:
So this all started when I ran into a lot of loading issues (not just the above screen shot) because it was having a hard time over my weak wifi signal (-90dBm) AP.  I have a stronger AP with the same SSID but SomnoTrace didn't use that one.  That's why I have been improving the wifi selection strategy....I have a few more fixes coming, that's why I put this on draft for now.

ilyakruchinin:
The first thing in fixing something always comes from identifying the problem.
The first question to ask is "which problem are we trying to solve".
The second question is "are we confident that this is a problem worth fixing, are we certain we understand the problem itself, and do we fully understand the underlying root cause".

I get the impression that this PR was based on incorrect understanding, thus the goals of the PR might not be beneficial. Please reassess the rationale behind the PR.

ilyakruchinin:
## PR #141 Assessment

### Motivating Evidence: Invalid

The 502 error screenshot is from **OpenResty** reverse proxy — not SomnoTrace or ESP32. The PR is solving a problem that wasn't demonstrated to be SomnoTrace's fault.

### Changes

**1. BSSID/channel passed to supplicant — real bug fix, over-engineered**

The current scan loop already finds the strongest AP per SSID and stores it in `cands[i].rec`. But [try_single_ssid()](cci:1://file:///opt/shared/SomnoTrace/main/net_provision.c:448:0-495:1) only receives the SSID string — it discards the BSSID and channel, letting the supplicant grab an arbitrary (possibly weaker) BSS. The fix is ~3 lines: copy `rec.bssid`, set `bssid_set`, set `channel`. The PR wraps this in a new parameter, attempt-conditional pin/fallback logic, and logging — unnecessary complexity for a dropped variable.

**2. Authmode threshold `WIFI_AUTH_OPEN` → `WIFI_AUTH_WPA2_PSK` — regression**

The current `WIFI_AUTH_OPEN` threshold means "accept any auth mode." The PR changes it to `WIFI_AUTH_WPA2_PSK` when a password is set, which **rejects WPA3-SAE and WPA3-only networks**. As WPA3 adoption grows (it's been mandatory on new routers certified since 2020), this would prevent SomnoTrace from connecting to increasingly common modern networks. The PR's stated reason — "silences the supplicant's auto-raise warning" — is cosmetic; the auto-raise is correct behaviour.

**3. Removing cross-slot RSSI sorting — regression**

Current code sorts all visible configured networks by signal strength and tries the strongest first. The PR replaces this with strict slot priority: Network #1 is always tried first, regardless of signal. If the user is at a different location where Network #1 is unreachable but Network #2 is strong, the device wastes **3 attempts × 5s spacing = 15+ seconds** failing on Network #1 before even trying #2. For a portable device used in multiple locations (home, travel, hospital), this is a significant usability regression.

**4. Scan cap 32 → 60 — likely ineffective**

`esp_wifi_scan_get_ap_records()` is bounded by the scan buffer allocated in `WIFI_INIT_CONFIG_DEFAULT()`. Increasing the application-side cap without increasing `scan_cache_buf_size` likely returns the same number of records — a no-op disguised as a fix.

**5. PMF capable — fine**

`pmf_cfg.capable = true` is safe future-proofing. Low risk.

**6. Portal UI changes — fine**

Signal strength circles, connected markers, and `escHtml()` are good quality-of-life and security improvements. Could be merged independently.

### Verdict

Do not merge. One real bug fix (discarded BSSID) buried under two regressions (authmode, slot ordering), invalid evidence, and unnecessary complexity. The BSSID fix is a 3-line change that can be done independently.

ilyakruchinin:
`So this all started when I ran into a lot of loading issues (not just the above screen shot) because it was having a hard time over my weak wifi signal (-90dBm) AP. I have a stronger AP with the same SSID but SomnoTrace didn't use that one. `

Let's concentrate on this exact issue and fix it without over-engineering (tailored fix). A proper understanding of the root cause (exact code bug) behind this is the first starting point.

moldbuster:
Thanks for pointing me in the right direction and for the guidance.  I am def learning a lot here. You are 100% right to aim for a precise and minimal fix for the root causes. I'll break out them out into individual PRs.

PR #144, opened 2026-08-25
Animated logo + Screenshot API

moldbuster:
## Summary

- **Animated logo on screen**: brings the web UI's animated logo onto the status screen as a 36 px procedural emblem beside the title. A logical animation clock slows down instead of jumping when renders fall behind, keeping motion smooth (`main/bsp_display.c`)
- **ESP32 Screen capture API (for documentation?)**: `GET /api/screenshot` serves the live framebuffer as a 24-bit BMP. The copy is made by the render task (sole framebuffer owner), so captures are tear-free even during 25 Hz graph streaming (`main/net_provision.c`, `main/bsp_display.{c,h}`)

image

## Checklist

- [ X] This change is a clean-room implementation; no third-party copyrighted code has been copied without approval and notice in `THIRD-PARTY-NOTICES.md`.
- [ X] Any new source files include the standard license header from `docs/source-header.txt`.
- [ X] The code compiles and builds cleanly (`./scripts/build-dist.sh`).
- [ X] No real patient / medical data is included in test fixtures, commits, or descriptions.

### DIFF
--- a/main/bsp_display.c
+++ b/main/bsp_display.c
@@ -33,6 +33,7 @@
 #include "esp_log.h"
 #include "esp_attr.h"
 #include "esp_heap_caps.h"
+#include "esp_timer.h"
 #include "driver/gpio.h"
 #include "driver/spi_master.h"
 #include "esp_lcd_panel_io.h"
@@ -84,6 +85,8 @@ static void render_graph(void);
 static void render_status(void);
 static void display_task(void *arg);
 static void apply_panel_rotation(uint8_t degrees);
+static void logo_init(void);
+static void logo_render_frame(int cx, int cy);
 
 static esp_lcd_panel_handle_t s_panel = NULL;
 static esp_lcd_panel_io_handle_t s_io = NULL;
@@ -123,7 +126,13 @@ typedef enum {
 #define STATUS_LINE_LEN    48
 
 #define DISPLAY_TASK_STACK 4096
-#define STATUS_FRAME_MS    1000  /* status refresh cadence (live RSSI) */
+/* Status refresh cadence. Doubles as the frame clock for the animated brand
+ * badge (10 fps — deliberately modest; the therapy graph streams at 25 Hz).
+ * The badge's phase follows a logical clock that never advances more than
+ * ~1.25 frames at a step, so a late or interrupted render slows the
+ * animation briefly instead of letting it jump/stutter. Raise this to save
+ * power at the cost of smoothness. */
+#define STATUS_FRAME_MS    100
 
 /* ── Flow graph layout (static, non-adaptive) ───────────────────────────
  * GRAPH_FULL_SCALE is the fixed full-scale deflection (L/min) from the zero
@@ -161,6 +170,15 @@ static float *s_flow_local;   /* render-task snapshot */
 static float *s_flow_yf;      /* render-task y coordinates */
 static int    s_flow_head = 0;
 static int    s_flow_count = 0;
+
+/* Screenshot support. The framebuffer has a single owner (the render task),
+ * so captures are requested via s_snap_want and performed inside
+ * display_task; callers read the finished copy from s_snap_buf. All three
+ * flags are guarded by s_state_mutex. */
+static uint16_t *s_snap_buf = NULL;          /* module-owned snapshot buffer */
+static SemaphoreHandle_t s_snap_sem = NULL;  /* given when a copy completes */
+static bool s_snap_want = false;             /* copy requested, not yet done */
+static bool s_snap_busy = false;             /* caller is reading s_snap_buf */
 
 /* ── Public state-mutating API (never draws; render task handles drawing) ── */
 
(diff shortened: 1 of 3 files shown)


ilyakruchinin:
What's the reason for a new API call to GET /api/screenshot - which problem does it solve?

moldbuster:
The screen shot API is a temporary developer feature to get clean screen shots for documentation.  Used it to capture the screen shot in this PR.  Is there any other way to get a clean screen shot of the ESP32 screen?  If this is not needed any more, I can certainly remove it.

ilyakruchinin:
OK, I'm not quite happy with the implementation and the goals of this PR in general.
Animated logo is cool, but displaying it shouldn't become a burden on CPU/RAM/SPI bus.

Currently, I am seeing this:
Device Telemetry
Int. Free
31.0 KB
Int. Min
7.9 KB
Int. LFB
12.5 KB

Which means we're approaching critical values on internal RAM available.
I have already received a report from a user on OOM event when trying to get OTA update.

I have also seen reports of SPI bandwidth issues and LCD screen freezing (unrelated to this PR, but worth noting that SPI needs to be properly controlled/managed).

In short: we're very short on internal RAM (pun intended).

## PR #144 Review: Animated Logo + Screenshot API

### Problems

1. **STATUS_FRAME_MS 1000→100ms**: 10x more full-screen re-renders + SPI flushes in status mode (1.15 MB/s vs 115 KB/s). Unnecessary — only the 36×36 logo region changes between frames.

2. **112 KB PSRAM permanently allocated** for `s_snap_buf` (screenshot buffer) at init — even though the author confirmed it's a temporary dev feature for documentation. No access control on the endpoint either.

3. **~1.4 KB internal RAM** consumed by static scratch arrays (`s_logo_y_main[180]`, `s_logo_y_echo[180]`, `s_logo_col_ok[180]`). Device already shows only 7.9 KB internal free — this makes it worse.

4. **Per-frame FPU cost**: `sinf()`/`cosf()`/`sqrtf()` across ~200+ pixels every 100ms. Unnecessary on a 240 MHz ESP32-S3 with limited FPU throughput.

5. **Full-screen fb_clear() + render_status every 100ms** even when nothing changed except the logo animation phase.

### Better Approach

- **Partial region refresh**: Keep `STATUS_FRAME_MS` at 1000ms. Add a separate 5 Hz logo timer that only re-renders and flushes the 36×36 logo sub-region via `esp_lcd_panel_draw_bitmap(panel, x, y, x+36, y+36, ptr)`. SPI bandwidth drops from 1.15 MB/s to ~6.5 KB/s.

- **Remove screenshot API** (or allocate on-demand / gate behind compile flag). Frees 112 KB PSRAM.

- **Pre-bake animation frames** at init: render 30 frames into PSRAM (~78 KB, one-time FPU cost). Runtime becomes a simple `memcpy` of 2.6 KB per frame — zero FPU.

- **Move scratch arrays to PSRAM** or eliminate them with pre-baked frames. Recovers ~1.4 KB internal RAM.

**Net result**: ~0 KB internal RAM impact, ~82 KB PSRAM (sprite + frames), ~6.5 KB/s SPI, negligible CPU. vs PR's ~1.4 KB IRAM, ~116 KB PSRAM, 1.15 MB/s SPI, heavy FPU.

ilyakruchinin:
I'd recommend the following:
1) maybe wait till I investigate further options to free up internal RAM (I've already done it twice, but every time it gets more challenging - it is planned work, I definitely WILL do it soon).
2) the animation part of the PR - have a look at the "better approach" section above. It achieves the same fancy effect, with with way less burden on the CPU/RAM/SPI. Maybe as a separate PR (no guarantees though, see the RAM note above).
3) the screenshot API part - generally OK, except for the constant 112KB PSRAM reserved block. Honestly, I am not convinced we need it (it's a nice "feature", but how often will we use this for the 112KB PSRAM waste?).

ilyakruchinin:
I have revisited t this PR/task.
As I mentioned earlier, the PR "as is" right now adds a lot of overhead on SPI (and a bit on RAM/CPU).
I am a bit worried about the potential implications, as I've already seen and received reports about LCD instability (which is something I'm working on at the moment). I've even tried the "partial refresh approach" that I proposed earlier with glitchy results (my own testing, not saying it's not possible, but my "initial attempt" wasn't a good one).

I did, however, trialled a screenshot API with dymanic on-demand PSRAM allocation (the 112KB mentioned earlier).
**It worked well for everything except the live graph**.

Having thought about it, here's what I've decided.
1) animated logo on the main screen - raises code complexity while increasing SPI bandwidth with no measurable benefit (let's be honest, people don't really look at the LCD all day to observe the same animated logo - even the Live Therapy Graph, while questionable, does carry certain real meaning - an actual live waveform, but a logo does not). I've decided to **NOT** implement the animated logo on the LCD.
2) screenshot API - happy to go ahead with dynamic PSRAM allocation, but code would need to be separated from the animated logo (into a different PR). Feel free to resubmit "just the screenshot API" PR separately (but ensure to dynamically allocate PSRAM on screenshot request).

As a result, I'm closing this PR as "won't implement".
Thank you for suggesting innovations for SomnoTrace and helping make it better!

PR #161, opened 2026-08-29
fix(sd_storage): allow format of exFAT cards; pre-erase before f_mkfs

cwm12345:
## Problem

Two bugs prevent SomnoTrace from reformatting a card that carries an exFAT filesystem (the default on any 64 GB SDXC card inserted on Windows or macOS).

**Bug 1 — format button silently does nothing on mount failure**

sd_storage_format() guards on !s_mounted || !s_card, but s_card is only assigned after a *successful* VFS mount. When a card with an unrecognised filesystem (exFAT) is inserted, mount fails and s_card stays NULL. Tapping 'reformat' then returns ESP_ERR_INVALID_STATE immediately — the user sees the button do nothing but reports success even though the card is fully detected at the SDMMC layer.

**Bug 2 — f_mkfs leaves a non-mountable volume when exFAT signatures are present**

esp_vfs_fat_sdcard_format() calls _mkfs() without zeroing the disk first. An exFAT card leaves a residual MBR

### DIFF
--- a/main/sd_storage.c
+++ b/main/sd_storage.c
@@ -301,20 +301,67 @@ void sd_storage_lease_release(sd_lease_t role)
 
 esp_err_t sd_storage_format(void)
 {
-    if (!s_mounted || !s_card) {
-        ESP_LOGE(TAG, "format: SD card not mounted");
-        return ESP_ERR_INVALID_STATE;
-    }
-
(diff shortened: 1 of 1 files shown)


cwm12345:
I have read the CLA Document and I hereby sign the CLA

ilyakruchinin:
Thank you for your contribution to SomnoTrace.
Good improvement, merged!

PR #166, opened 2026-08-30
Feature/wifi fix

moldbuster:
Multiple wifi-related issues caused problems.  This PR serves as a **proof of concept**.  Feel free to use it however is best.

### **Issue 1: BLE Initialization Interruption Prevents Connection to Primary Wi-Fi (wifi 1)**

* **What is the problem?**
  When BLE initializes concurrently with Wi-Fi provisioning, radio time-slot contention on single-radio hardware (ESP32-S3) causes Wi-Fi probe/association timeouts on saved primary credentials (wifi 1: `boulangerie`). 

  Because BLE auto-reconnection occupies radio time slices during startup, wifi 1 fails to associate within the initial connection window. The system does not report BLE packet loss or driver errors; instead, `netprov` silently times out on wifi 1 after 5 seconds and falls back to secondary credentials (wifi 2: `patisserie`):

Logs:
```
  I (15:11:59.169) as11_ble: BLE initialised                                       <-- 1. BLE stack fully ready
  I (15:11:59.171) as11_ble: saved pairing found, auto-reconnecting...              <-- 2. BLE auto-reconnect starts
  I (15:11:59.172) as11_ble: reconnect: connecting to E0:79:8D:A4:5E:CC (ResMed 924935) <-- 3. Active BLE scan/connection
  ...
  I (15:11:59.255) somnotrace:   Connecting to Wi-Fi...                            <-- 4. Wi-Fi connection phase begins
  I (2046) wifi:mode : sta (28:84:85:49:8e:ec)                                     <-- 5. wifi 1 ('boulangerie') attempted
  I (2046) wifi:enable tsf                                                         <-- 6. RF contention skips wifi 1
  ...
  I (15:12:04.271) netprov: trying candidate 1: 'patisserie' (-54 dBm)             <-- 7. 5s timeout forces fallback to wifi 2
  ...
  I (15:12:05.340) netprov: connected to 'patisserie', ip=192.168.10.162          <-- 8. wifi 2 successfully connected
```

* **Justification** 

In testing, initializing BLE before Wi-Fi caused consistent Wi-Fi probe/handshake timeouts on the primary network (wifi 1: boulangerie), forcing a fallback to the secondary network (wifi 2: patisserie).

Reordering main.c so Wi-Fi provisions before NimBLE starts resolved this deterministic timeout in testing by giving Wi-Fi uninterrupted radio access during its CONNECTING phase. According to Espressif's documentation, the single radio chip rely on Time Division Multiplexing (TDM) across distinct states (IDLE, SCAN, CONNECTING, CONNECTED). Deferring BLE startup keeps Bluetooth in an IDLE state during Wi-Fi setup, granting Wi-Fi 100% radio availability during its initial WPA2 handshake.

As a proof of concept, this empirical result confirms that startup sequencing directly impacts the wifi 1 connection issue and allows wifi 1 to connect consistently on boot. Additionally, updating netprov_try_connect() to fall back to driver default scan parameters while BLE is active prevents scan window collisions during subsequent runtime reconnects.

* **Files Changed**
  * `main.c`
  * `net_provision.c`

---
### **Issue 2: Settings Page: Masked password handling & Placeholder UX**

* **What is the problem?**
  Previously, presaved passwords were not visually indicated in the form, leaving users unclear on whether credentials were stored. 

* **Justification**
  1. **Masked Pre-Fill:** Saved networks load with a masked bullet placeholder (`••••••••`), giving immediate visual confirmation that credentials are stored.
  2. **Auto-Clear on Focus (`onPassFocus`):** Clicking or focusing into the password input field now automatically clears the placeholder dots before typing begins. This allows users to immediately enter a new password or leave the field empty for open networks without editing placeholder characters.
  3. **Open Network Guidance:** Updated empty field placeholder text to: "Leave this empty for an open network, otherwise provide 8+ characters", which clearly guides the user on proper input formatting.

Existing password is obvious from masked placeholder:
image

Empty password (open network) visible tip:
image

* **Files Changed**
  * `portal.html`
 
---

### **Issue 3: Missing WPA2 Length Validation and Human Readable Wifi Disconnect Logs**

* **What is the problem?**
  Previously, submitting password lengths between 1 and 7 characters bypassed both client side and server side checks. This resulted in invalid short credentials being written to storage. 

  During bootup, the ESP-IDF Wi-Fi driver attempts association but fails the 4-way WPA2 key exchange due to the invalid key length. This triggers immediate state tear-downs (`run -> init (0xf00)`) and leaves the network stack stuck in an endless retry loop without clear diagnostic logs:

  ```text
  I (16:25:48.195) netprov: trying candidate 1: 'patisserie' (-55 dBm)
  I (7066) wifi:state: init -> auth (0xb0)
  I (7076) wifi:state: assoc -> run (0x10)                   <-- 1. Associated with AP
  I (10116) wifi:state: run -> init (0xf00)                  <-- 2. 4-way WPA2 handshake fails due to invalid key
  I (10126) wifi:Coexist: Wi-Fi connect fail, apply reconnect coex policy <-- 3. Driver tear-down
  I (16:25:51.277) netprov: retry connect (1/3)              <-- 4. Retries repeatedly
  ...
  W (16:26:43.217) netprov: all candidates exhausted         <-- 5. All retry attempts fail
Justification

* Client-Side Portal Guard (portal.html): Updated validateForm() to strictly enforce IEEE 802.11i WPA2 constraints (8 to 63 characters or empty for open networks). Submitting 1 to 7 characters highlights the field in red, displays an explicit message ("Fix password length (8+ chars or leave empty for open)"), and disables the save button.
* Server-Side Validation & NVS Guard (net_provision.c): Added length checking in save_post_handler() to reject invalid 1 to 7 character payloads before executing NVS flash writes, followed by post-write readback verification.
* Human-Readable Disconnect Diagnostics: Implemented wifi_disconnect_reason_str() mapping to convert raw ESP-IDF disconnect reason codes into human-readable log strings, providing clear feedback during connection drops.

Invalid password (red highlight):
<img width="778" height="582" alt="image" src="https://github.com/user-attachments/assets/8c058b22-d91b-4d2b-a661-0bea948d8c9d" />

* **Files Changed**
  * `net_provision.c`
  * `portal.html`

---

## Checklist

- [ X] I have read and agree to the [Contributor License Agreement (CLA)](CLA/individual-cla.md) (or [Corporate CLA](CLA/corporate-cla.md)).
- [ X] This change is a clean-room implementation; no third-party copyrighted code has been copied without approval and notice in `THIRD-PARTY-NOTICES.md`.
- [ X] Any new source files include the standard license header from `docs/source-header.txt`.
- [ X] The code compiles and builds cleanly (`./scripts/build-dist.sh`).

### DIFF
--- a/main/main.c
+++ b/main/main.c
@@ -189,8 +189,23 @@ void app_main(void)
         ESP_LOGW(TAG, "audio codec init failed — buzzer will be unavailable");
     }
 
-    /* 4c. Initialise BLE (AirSense 11 pairing). Non-fatal on failure.
-     * Runs after storage init + crash recovery (see 4b). */
+    /* 4c. Try to connect to configured Wi-Fi BEFORE BLE init.
+     * This allows custom active scan params (20ms/channel) to be honoured.
+     * When BLE is already running, ESP-IDF forces BT-coexistence-safe scan
+     * params and ignores our active scan dwell time. */
+    struct netprov_config cfg;
+    bool has_creds = netprov_load_config(&cfg);
+
+    char ip[16] = "0.0.0.0";
+    esp_err_t err = ESP_FAIL;
+    if (has_creds) {
+        const char *lines[] = { "Connecting to Wi-Fi..." };
+        show_status("SomnoTrace", lines, 1);
+        err = netprov_try_connect(&cfg, ip, 35000);
+    }
+
+    /* 4c-bis. Initialise BLE (AirSense 11 pairing). Non-fatal on failure.
+     * Runs after Wi-Fi connect attempt so active scan params are honoured. */
     if (as11_ble_init() != ESP_OK) {
         ESP_LOGE(TAG, "BLE init failed; CPAP pairing unavailable");
     }
@@ -203,7 +218,7 @@ void app_main(void)
         ESP_LOGE(TAG, "Oximeter init failed; O2 Ring sync unavailable");
     }
 
-    /* 4c-bis. BLE startup has begun, so reconnect can now establish whether
+    /* 4c-quater. BLE startup has begun, so reconnect can now establish whether
      * therapy is already running.  Only now is it safe to let the idle post
      * worker export days that boot recovery queued: doing it earlier could
      * run a multi-minute rebuild while a live session was trying to start. */
@@ -214,26 +229,13 @@ void app_main(void)
     therapy_alert_set_therapy_active_fn(bsp_display_is_therapy_active);
     therapy_alert_init();
 
-    /* 5. Load config from NVS. */
-    struct netprov_config cfg;
-    bool has_creds = netprov_load_config(&cfg);
-
-    /* 6. If BOOT was held at boot, force SoftAP regardless. */
+    /* 5. If BOOT was held at boot, force SoftAP regardless. */
     if (s_softap_requested) {
         ESP_LOGW(TAG, "BOOT long-press detected: forcing SoftAP");
         enter_softap(&cfg);
         while (true) {
             vTaskDelay(pdMS_TO_TICKS(1000));
         }
-    }
-
-    /* 7. Try to connect to configured Wi-Fi. */
-    char ip[16] = "0.0.0.0";
-    esp_err_t err = ESP_FAIL;
-    if (has_creds) {
-        const char *lines[] = { "Connecting to Wi-Fi..." };
-        show_status("SomnoTrace", lines, 1);
-        err = netprov_try_connect(&cfg, ip, 15000);
     }
 
     bool in_softap = false;
--- a/main/net_provision.c
+++ b/main/net_provision.c
@@ -85,6 +85,43 @@ static const char *TAG = "netprov";
 #define WIFI_FAIL_BIT       BIT1
 #define MAX_STA_RETRY       3
 
+/* Map Wi-Fi disconnect reason codes to human-readable strings. */
+static const char *wifi_disconnect_reason_str(wifi_err_reason_t reason)
+{
+    switch (reason) {
+    case WIFI_REASON_UNSPECIFIED:                  return "unspecified";
+    case WIFI_REASON_AUTH_EXPIRE:                  return "auth_expire";
+    case WIFI_REASON_AUTH_LEAVE:                   return "auth_leave";
+    case WIFI_REASON_ASSOC_EXPIRE:                 return "assoc_expire";
+    case WIFI_REASON_ASSOC_TOOMANY:                return "assoc_toomany";
+    case WIFI_REASON_NOT_AUTHED:                   return "not_authed";
+    case WIFI_REASON_NOT_ASSOCED:                  return "not_assoced";
+    case WIFI_REASON_ASSOC_LEAVE:                  return "assoc_leave";
+    case WIFI_REASON_ASSOC_NOT_AUTHED:             return "assoc_not_authed";
+    case WIFI_REASON_DISASSOC_PWRCAP_BAD:          return "pwrcap_bad";
+    case WIFI_REASON_DISASSOC_SUPCHAN_BAD:         return "supchan_bad";
+    case WIFI_REASON_IE_INVALID:                   return "ie_invalid";
+    case WIFI_REASON_MIC_FAILURE:                  return "mic_failure";
+    case WIFI_REASON_4WAY_HANDSHAKE_TIMEOUT:       return "4way_handshake_timeout";
+    case WIFI_REASON_GROUP_KEY_UPDATE_TIMEOUT:     return "group_key_update_timeout";
+    case WIFI_REASON_IE_IN_4WAY_DIFFERS:           return "ie_in_4way_differs";
+    case WIFI_REASON_GROUP_CIPHER_INVALID:         return "group_cipher_invalid";
+    case WIFI_REASON_PAIRWISE_CIPHER_INVALID:      return "pairwise_cipher_invalid";
+    case WIFI_REASON_AKMP_INVALID:                 return "akmp_invalid";
+    case WIFI_REASON_UNSUPP_RSN_IE_VERSION:        return "unsupp_rsn_ie_version";
+    case WIFI_REASON_INVALID_RSN_IE_CAP:           return "invalid_rsn_ie_cap";
+    case WIFI_REASON_802_1X_AUTH_FAILED:           return "802_1x_auth_failed";
+    case WIFI_REASON_CIPHER_SUITE_REJECTED:        return "cipher_suite_rejected";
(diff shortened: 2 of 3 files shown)


ilyakruchinin:
I hsv

ilyakruchinin:
I have not yet checked the PR, but can immediately see it's statement is incorrect on placeholders.
If you don't change anything in the password field, the "Save and Reboot" button is not active.


Please re-validate the statements in the PR.

ilyakruchinin:
Screenshot_20260831-074105

ilyakruchinin:
This statement:
"Initializing Bluetooth before Wi-Fi causes the active scan configuration (20ms dwell time per channel) to collide with Bluetooth operations, stalling Wi-Fi connection attempts or triggering scan failures"

Directly contradicts the vendors statement on WiFi/BLE coexistence. Is there any substantial proof to this statement? Log messages? Measurements?

Looks like an AI hallucination.

ilyakruchinin:
"Submitting password lengths between 1 and 7 characters caused obscure driver and NVS write failures. Additionally, the web portal did not visually guide users on open network setups, and Wi-Fi disconnect events lacked actionable log output."

Is there definitive proof/evidence to NVS write failures? Please provide.

ilyakruchinin:
All 3 statements are not substantiated by any evidence and contradict what I'm experiencing.

Please re-check and re-validate each statement and corroborate with real evidence.

moldbuster:
@ilyakruchinin I've updated this PR, hopefully it makes more sense now.  Thanks.

ilyakruchinin:
There is no such thing as "primary WiFi"

WiFi order does not matter. I think you're going down a rabbit hole on this assumption (and all the false conclusions based on this assumption).
All WiFi slots are equal. They ONLY matter in case of same signal strength.
But if signal strength is NOT the same -> they are evaluated based on it ONLY (not by slot order, simply by choosing WiFi SSID with the strongest signal).

Could you please run a research on how WiFi selection is handled? Thanks.

ilyakruchinin:
Hi @moldbuster, thanks for the contribution. I've reviewed the PR in detail and have some concerns. Let me address each of the three issues separately.

## On combining unrelated changes

Before getting into specifics: bundling three unrelated "fixes" into a single PR makes review significantly harder and prevents selective merging. If one part has problems, the whole PR gets blocked. In future, please split unrelated changes into separate PRs — one concern per PR.

## Issue 1: BLE/WiFi init ordering — not accepted

I don't believe the premise holds up, and the proposed change introduces a regression for offline use.

**The scan-params argument is incorrect.** The PR description states that "when BLE is already running, ESP-IDF forces BT-coexistence-safe scan params and ignores our active scan dwell time." But `netprov_try_connect()` (the boot-time connect path) uses a zero-initialized `wifi_scan_config_t` with no custom dwell time — it never set active scan params in the first place. The 20ms/channel active scan configuration exists only in `wifi_scan_task()`, which is a separate function used for the SoftAP WiFi scan list, and that function already correctly checks `s_portal_mode` and passes NULL to `esp_wifi_scan_start` when BLE is active.

So the change to `netprov_try_connect()` (passing NULL instead of `&scan_cfg` when not in portal mode) is a no-op — both paths already use driver-default parameters.

**The logs don't demonstrate coexistence failure.** The provided logs show normal SSID fallback from "boulangerie" to "patisserie." There are no BLE coexistence errors, no scan failures, no RF contention warnings from the driver. The annotations interpreting the logs as "RF contention skips wifi 1" are not supported by any driver-level message. The fallback behavior shown is consistent with normal candidate-by-RSSI selection when the strongest SSID doesn't complete association within the per-SSID timeout.

This also directly contradicts Espressif's documented WiFi/BLE coexistence design, where TDM coexistence is a supported and managed configuration on the ESP32-S3.

**The regression.** The current boot sequence starts BLE first, then attempts WiFi (15s timeout). If WiFi is unavailable (camping, travel, router down), BLE is already connecting during that window, so the device reaches usable "degraded mode" state in ~15-20 seconds.

The PR reverses this: WiFi attempts first with a **35-second timeout** (more than doubled), then BLE starts, then the degraded-mode logic waits another 30 seconds for BLE to connect. This puts offline/camping users at ~65+ seconds before the device is usable — a 3x+ regression for a use case that the device is specifically designed to support.

## Issue 2: Password placeholder UX — not accepted as described

The PR states that "presaved passwords were not visually indicated in the form." This is already implemented. The current code renders `••••••••` as the placeholder for stored passwords and `Password` for empty slots, giving the visual confirmation the PR claims to add.

Additionally, the Save button is already disabled when no fields have been modified, so the concern about users being "unclear on whether credentials were stored" doesn't apply to the current implementation.

The minor UX polish (auto-clear on focus, "Leave this empty for an open network" hint text) is reasonable but should be a small standalone PR if you'd like to pursue it — not bundled with WiFi stack changes.

## Issue 3: WPA2 length validation and disconnect diagnostics — partially accepted

This is the one with a real basis. The current code has no password length validation on either the client or server side, so a 1-7 character password can be saved to NVS and will cause 4-way handshake failures with pointless retries.

The following parts are good fixes that I'd accept as a standalone PR:
- Server-side length check rejecting 1-7 character passwords (HTTP 400)
- Client-side validation with visual feedback
- Fast-fail on auth-related disconnect reasons instead of retrying 3x
- Human-readable disconnect reason strings in logs

The verify-after-write readback (reading NVS config back and comparing passwords after save) is unnecessary complexity — NVS writes are reliable, and this adds code path surface for no practical benefit.

If you'd like to resubmit Issue 3's core fixes as a focused standalone PR, I'd be happy to review and merge it.

moldbuster:
Thanks @ilyakruchinin for going over this with me.  I really appreciate it.  To wrap up here:

**Issue 1** 

Yeah, even though the BLE and wifi did clash during startup in my testing, I agree keeping the boot time faster for camping mode is the right call.  I'm with you on that.

And just for the record: I did not change any of the wifi selection logic (primary/secondary priority etc.) in any of these 3 issues.  The fixes here had nothing to do with the wifi selection logic.

**Issue 2**

It is worth clarifying the UX intent here to highlight where our perception differs:

The current code relies on an HTML placeholder="••••••••", which leaves the input field completely empty in the DOM (value = ""). Because it is only background placeholder text, clicking or focusing into the box leaves the dots rendered under the cursor until typing begins. In practice, this confuses users into thinking they have to backspace over those dots before entering a new key.

The goal was to separate passive visual hints from active form state and interaction handling.

Spawned PR: #170 

**Issue 3**

Spawned PR: #171

ilyakruchinin:
Thank you @moldbuster . I'll have a look a tomorrow or the day after (sorry quite busy at the moment).
Special thanks for the O2 calendar dot PR, I'll look into it, too.

moldbuster:
lol please take your time, there is no rush! You deserve a long vacation for everything you've done already =)

PR #170, opened 2026-08-31
Masked password handling & Placeholder UX

moldbuster:
Spawned from #166 

It is worth clarifying the UX intent here to highlight where our perception differs:

The current code relies on an HTML placeholder="••••••••", which leaves the input field completely empty in the DOM (value = ""). Because it is only background placeholder text, clicking or focusing into the box leaves the dots rendered under the cursor until typing begins. In practice, this confuses users into thinking they have to backspace over those dots before entering a new key.

This PR used the sentinel value (PW_SENTINEL_WIFI), dynamic focus clearing and adaptive helper text so the field explicitly clears on focus and provides a better open network setup. The goal was to separate passive visual hints from active form state and interaction handling.

### **Issue 2: Settings Page: Masked password handling & Placeholder UX**

* **What is the problem?**
  Previously, presaved passwords were not visually indicated in the form, leaving users unclear on whether credentials were stored. 

* **Justification**
  1. **Masked Pre-Fill:** Saved networks load with a masked bullet placeholder (`••••••••`), giving immediate visual confirmation that credentials are stored.
  2. **Auto-Clear on Focus (`onPassFocus`):** Clicking or focusing into the password input field now automatically clears the placeholder dots before typing begins. This allows users to immediately enter a new password or leave the field empty for open networks without editing placeholder characters.
  3. **Open Network Guidance:** Updated empty field placeholder text to: "Leave this empty for an open network, otherwise provide 8+ characters", which clearly guides the user on proper input formatting.

Existing password is obvious from masked placeholder:
image

Empty password (open network) visible tip:
image

* **Files Changed**
  * `portal.html`
 
## Checklist

### DIFF
--- a/main/portal.html
+++ b/main/portal.html
@@ -1266,6 +1266,54 @@ function togglePw(id) {
   }
 }
 
+/* ── WiFi Password Field Tracking ──────────────────────────────── */
+function onPassInput(el) {
+  if (!el) return;
+  var isSentinel = el.value === PW_SENTINEL_WIFI;
+  if (isSentinel) {
+    el.value = '';
+  }
+  // Show hint immediately when field is empty
+  if (el.value === '') {
+    el.placeholder = 'Leave this empty for an open network, otherwise provide 8+ characters.';
+  }
+  el.setAttribute('data-touched', 'true');
+}
+
+function onPassKeydown(el, e) {
+  if (!el) return;
+  var isSentinel = el.value === PW_SENTINEL_WIFI;
+  // Backspace/Delete on sentinel -> clear entire field
+  if (isSentinel && (e.key === 'Backspace' || e.key === 'Delete')) {
+    e.preventDefault();
+    el.value = '';
+    el.placeholder = 'Leave this empty for an open network, otherwise provide 8+ characters.';
+  }
+}
+
+function onPassBlur(el) {
+  if (!el) return;
+  var len = el.value.length;
+  // Validate: 0 (open) or 8+ (WPA2) OK, 1-7 invalid
+  var wasInvalid = el.getAttribute('data-invalid') === 'true';
+  var nowInvalid = (len > 0 && len < 8);
+  if (nowInvalid) {
+    el.style.borderColor = 'var(--danger)';
+    el.style.boxShadow = '0 0 0 3px rgba(248,113,113,0.3)';
+  } else {
+    el.style.borderColor = '';
+    el.style.boxShadow = '';
+  }
+  if (wasInvalid !== nowInvalid) {
+    el.setAttribute('data-invalid', nowInvalid ? 'true' : 'false');
+    updateSaveButton();  // Re-check save button state
+  }
+  // Show hint when empty
+  if (el.value === '') {
+    el.placeholder = 'Leave this empty for an open network, otherwise provide 8+ characters.';
+  }
+}
+
 /* ── Dirty State Tracking ──────────────────────────────────────── */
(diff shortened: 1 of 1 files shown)


ilyakruchinin:
# PR #170 — Masked password handling & Placeholder UX

## Verdict: DO NOT MERGE — Critical password corruption bug

## Problem 1: Password corruption when editing an existing network's password

The PR sets the actual DOM [value](cci:1://file:///opt/shared/SomnoTrace/main/portal.html:4722:23-4722:91) of the password input to the sentinel string `█UNCHANGED█` (`PW_SENTINEL_WIFI`) when loading saved networks:

```javascript
passEl.value = PW_SENTINEL_WIFI;
```

The `onPassInput()` handler is supposed to detect and clear the sentinel when the user interacts with the field:

```javascript
function onPassInput(el) {
  var isSentinel = el.value === PW_SENTINEL_WIFI;
  if (isSentinel) {
    el.value = '';
  }
  ...
}
```

**Why this fails:** The `input` event fires *after* the browser has already inserted the typed character into the field. If the user focuses the field and types `s`, the DOM value becomes `█UNCHANGED█s` (24 chars) before `onPassInput` runs. The equality check `el.value === PW_SENTINEL_WIFI` evaluates to **false** because the value is now `█UNCHANGED█s`, not `█UNCHANGED█`. The sentinel is never cleared.

**Result:** The user's new password is saved as `█UNCHANGED█secret123` instead of `secret123`. The ESP32 attempts to connect with the corrupted string and permanently fails. This happens on **every** password edit where the user types into a field that had a stored password.

**How to verify:** Open the portal, load a network with a saved password, click into the password field, type a new password, and save. Check the HTTP request body — the password will contain the sentinel prefix.

## Problem 2: Sentinel string visible in plaintext via eye toggle

The existing [togglePw()](cci:1://file:///opt/shared/SomnoTrace/main/portal.html:1261:0-1274:1) function switches the input from `type="password"` to `type="text"`. Since the sentinel is now in the field's [value](cci:1://file:///opt/shared/SomnoTrace/main/portal.html:4722:23-4722:91) (not just the placeholder), clicking the eye icon displays `█UNCHANGED█` on screen in plaintext. This is confusing UX and leaks an internal implementation detail.

## Problem 3: Incomplete `data-touched` lifecycle

`onPassKeydown()` clears the sentinel on Backspace/Delete but does not set `data-touched="true"` or call [markWifiDirty()](cci:1://file:///opt/shared/SomnoTrace/main/portal.html:1301:0-1304:1), unlike `onPassInput()` which does both. After a save, `data-touched` is never reset to `"false"`, so a field touched in a previous edit session is treated as touched in the next session — breaking the sentinel-sending logic for unchanged passwords.

## Problem 4: `onPassBlur` doesn't mark the form dirty

If a user types 1-7 characters and blurs, the field is marked `data-invalid="true"` and the save button is disabled, but [markWifiDirty()](cci:1://file:///opt/shared/SomnoTrace/main/portal.html:1301:0-1304:1) is never called. If the user then fixes the length (types an 8th character via `onPassInput` which does call [markWifiDirty](cci:1://file:///opt/shared/SomnoTrace/main/portal.html:1301:0-1304:1)), it works — but if they clear the field and blur, the form remains clean despite having interacted with it.

## Recommendation

The existing pattern — `placeholder="••••••••"` with `value=""` — already communicates that a password is stored. If you want to improve the UX, use an `onfocus` handler to clear the field (before any input event) rather than putting the sentinel in the value. Track "touched" state with a data attribute set on `focus`, not `input`.

PR #173, opened 2026-08-31
fix(sd_storage): make "Format SD" work on 32-64 GB cards

cwm12345:
## Problem

"Format SD" in the maintenance panel currently does nothing on the card sizes it's meant to handle, and can panic the device.

**1. Silent failure on 32–64 GB cards.**
The format path's `allocation_unit_size` is `0`, which ESP-IDF defaults to the 512-byte sector size. On a 32–64 GB card that's ~60–130M FAT32 clusters and a FAT table hundreds of MB wide, built through a 4 KB work buffer — `f_mkfs` then returns `FR_MKFS_ABORTED` or runs long enough to trip the task watchdog. PR #161 had set `32768` for this; commit `d1747bd` reverted it and dropped the explanatory note.
The in-place path (card already mounted) had the same bug independently: `esp_vfs_fat_sdcard_format()` reuses `sd_storage_init()`'s mount config, whose `allocation_unit_size` is also `0`.

**2. The web UI always reported success.**
`/api/actions` returns `{"ok":true}` the moment `format_sd_task` is *spawned*, and `portal.html` then shows "SD card formatted. Rebooting…" with a countdown regardless of the outcome.

**3. Pressing Format during an upload scan panics the device.**
`format_sd_task` calls `uploader_reset_state()`, which wakes the upload scheduler to walk the card (`opendir`/`readdir`) at the same moment `sd_storage_format()` is in `f_unmount` + `f_mkfs` on that volume. The concurrent SDMMC traffic corrupts the driver:

```
sdmmc_read_sectors_dma ... returned 0x107 (timeout)
diskio_sdmmc: sdmmc_read_blocks failed (0x107)
assert failed: spinlock_acquire spinlock.h:142 (lock->count == 0)
Crashing task up_sched
```

The upload *pass* already takes `SD_LEASE_UPLOAD`; the *scan* never did.

**4. The post-format reboot was a hard reset with nothing flushed.**
`reboot_task` did `esp_restart()` ~1.5 s after `f_mkfs` with no `f_unmount` / `CTRL_SYNC`, so the reset could beat the card's own flush and the volume wouldn't mount until a second reboot.

## Changes

One commit, 6 files.

**`sd_storage.c` / `.h`**
- Add `SD_FORMAT_ALLOC_UNIT` (32 KB), used on both format paths, with a comment spelling out why `0` fails.
- In-place path: `esp_vfs_fat_sdcard_format()` → `esp_vfs_fat_sdcard_format_cfg()` with an explicit config.
- Drop the `slot.width = 1` retry: `esp_vfs_fat_sdmmc_mount()` leaves the VFS path registered on a post-registration failure, so the retry only returns `ESP_ERR_INVALID_STATE` and hides the real error. `sd_storage_init()` already does 4→1-bit for the normal mount.
- Clear `s_mounted` for the duration of the format so other subsystems' `sd_storage_is_ready()` checks report the card as away (stops `ox_canon` / `log_stream` erroring against the unmounted volume).
- Add `sd_storage_deinit()`: `f_unmount` (FAT sync + `CTRL_SYNC`) then release the handle. Called just before the deliberate reboot.

**`net_provision.c`**
- Add `s_format_progress` + `GET /api/format-progress` (mirrors the existing OTA-progress pattern). `format_sd_task` records `active` / `done` / `ok` / `error`.
- `format_sd_task` holds `SD_LEASE_DESTRUCTIVE` for the whole operation including the reboot, and refuses if the card is busy.
- On success: publish the result, wait 2.5 s for the browser poll, `sd_storage_deinit()`, then `esp_restart()`.
- Remove the `uploader_reset_state()` call (trigger for defect 3; the reboot re-inits the uploader from the empty state dir, and a failed format keeps the old state valid).
- Refuse a second format while one is running.

**`portal.html`**
- `format-sd` polls `/api/format-progress` and shows the reboot countdown, the real error string, or "SD busy — try again" only once the task is genuinely done. The confirmation dialog is unchanged.

**`upload_sched.c`**
- `do_scan()` and a new `reconcile_day_leased()` helper take the storage lease around every card walk — the same lease the upload pass already uses. A scan defers instead of racing a format (or a recording, or an export).

**`log_stream.c`**
- The flush task uses a sticky `s_sd_ready` latch; also check `sd_storage_is_ready()` so it stops writing while the card is away.

## Behaviour after this change

- Format on a 32/64 GB card (exFAT or FAT32): works in a few seconds, one clean reboot, card remounts empty on the first boot.
- Format while the uploader is mid-scan: `SD busy — try again`, succeeds on the retry a few seconds later.
- Format failure: portal shows `Format failed: <err>`, no reboot.
- No `up_sched` panic; no `ox_canon` / `log_stream` error spam during the format.

## Testing

Built with ESP-IDF v5.5.1, flashed to a Waveshare ESP32-S3-Touch-LCD-1.54. Serial on a clean run:

```
sd_storage: format: pre-erased first 8 sectors
vfs_fat_sdmmc: Formatting card, allocation unit size=32768
sd_storage: format: SD card formatted and directory tree recreated
sd_storage: deinit: SD flushed and unmounted
netprov: format_sd_task: rebooting for clean remount
[reboot]
sd_storage: SD mounted at /somnotrace, directory tree ready
```

Verified the `up_sched` panic no longer reproduces when Format is pressed during an active scan.

### DIFF
--- a/components/uploader/upload_sched.c
+++ b/components/uploader/upload_sched.c
@@ -554,6 +554,20 @@ static void run_pass(void)
 
 /* ── Task ─────────────────────────────────────────────────────────── */
 
+/* Reconcile one day's exported files against the index.  Holds the storage
+ * lease: the reconcile walks the card with opendir()/readdir(), and without
+ * the lease it can run while sd_storage_format() is calling f_mkfs on the same
+ * volume, which corrupts the SDMMC driver and panics this task. */
+static void reconcile_day_leased(uint32_t day)
+{
+    if (!uploader_lease_take(LEASE_WAIT_MS)) {
+        ESP_LOGD(TAG, "reconcile deferred: storage busy");
+        return;
+    }
+    upload_scan_reconcile_day(day);
+    uploader_lease_give();
+}
+
 static void do_scan(void)
 {
     int max_days = uploader_max_days();
@@ -572,6 +586,13 @@ static void do_scan(void)
         return;
     }
 
+    /* Same lease the upload pass takes — see reconcile_day_leased(). */
+    if (!uploader_lease_take(LEASE_WAIT_MS)) {
+        ESP_LOGD(TAG, "scan deferred: storage busy");
+        s_next_scan_us = now_us() + (int64_t)SCAN_INTERVAL_MS * 1000;
+        return;
+    }
+
     s_scanning = true;
     set_status("Scanning for new data");
     upload_scan_reconcile_all(max_days, slots, n_slots);
@@ -581,6 +602,7 @@ static void do_scan(void)
         free(ox_refs);
     }
     s_scanning = false;
+    uploader_lease_give();
     s_next_scan_us = now_us() + (int64_t)SCAN_INTERVAL_MS * 1000;
 }
 
@@ -612,7 +634,7 @@ static void sched_task(void *arg)
             switch (ev.type) {
             case EV_EXPORT:
                 ESP_LOGI(TAG, "export complete for %08u", (unsigned)ev.day);
-                upload_scan_reconcile_day(ev.day);
+                reconcile_day_leased(ev.day);
                 run_pass();
                 break;
 
@@ -620,7 +642,7 @@ static void sched_task(void *arg)
                 ESP_LOGI(TAG, "day %08u invalidated — will re-upload",
                          (unsigned)ev.day);
                 upload_index_forget_day(ev.day);
-                upload_scan_reconcile_day(ev.day);
+                reconcile_day_leased(ev.day);
                 run_pass();
                 break;
 
--- a/main/log_stream.c
+++ b/main/log_stream.c
@@ -206,6 +206,10 @@ static void log_flush_task(void *arg)
             ensure_log_dir();
             if (!s_sd_ready) continue;
         }
+        /* s_sd_ready is a latch; skip the flush while the card is actually
+         * away (e.g. a reformat in progress) so it does not log an error per
+         * tick against the unmounted volume. */
+        if (!sd_storage_is_ready()) continue;
 
         size_t avail;
         xSemaphoreTake(s_writebuf_mutex, portMAX_DELAY);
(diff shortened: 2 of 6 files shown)


ilyakruchinin:
Hmm.....
I thought zero (0) would automatically set it the the correct value regardless of partition size? Did it not work?

Because setting the block size to 32KB is a bit wasteful on smaller cards (each small fille will consume at least 32KB), so I was trying to avoid it.

I'll review this tomorrow or the day after (but a bit surprised that zero didn't work for "auto").

ilyakruchinin:
All merged, thank you for your contribution @cwm12345 !

ilyakruchinin:
Released in https://github.com/ilyakruchinin/SomnoTrace/releases/tag/v1.3.2
Could you please test? 
Thank you!!!

cwm12345:
Good question — I assumed the same. However, it doesn't appear that 0 reaches FatFs's auto-sizing in ESP-IDF: esp_vfs_fat_sdcard_format_cfg() runs the value through esp_vfs_fat_get_allocation_unit_size(), which does MAX(requested, sector_size), so 0 becomes 512 and f_mkfs never gets the 0 that would trigger its volume-based default. The serial log shows it: Formatting card, allocation unit size=512. On a 16–64 GB card that's ~31–125M clusters / 120–500 MB FAT through a 4 KB buffer.

Agreed that a flat 32 KB is heavy for smaller cards. We could switch it to a capacity-scaled pick (4 K ≤8 GB, 8 K ≤16 GB, 16 K ≤32 GB, 32 K above — matching Windows' FAT32 table), keyed off card->csd.capacity. However, I recently shopped for SD Cards for this project and had trouble finding even a 16GB card so I wonder how much practical use this would have? Say the word and I can work on that.

cwm12345:
Tested v1.3.2 and looks good in my environment!

ilyakruchinin:
Awesome!
Let's not pursue perfection, 32KB block is fine (and, like you mentioned, the SD card sizes are massive nowadays, no need to try saving every kilobyte per file). I am happy with what we have.

Thank you for your contribution and making SomnoTrace better!

PR #179, opened 2026-09-01
fix(sd_storage): use opendir() for the post-format-failure card probe

cwm12345:
Turned up in a post-merge review of #173's follow-up commit (`9eaa72d`).

## Problem

The probe that restores `s_mounted` when the card survived a failed format uses `fopen(SD_APP_DIR, "r")`. `SD_APP_DIR` is a directory, and `fopen("…","r")` on a directory returns NULL on the FATFS VFS even when the card is mounted — so `s_mounted` is never restored and the branch is effectively dead (a failed format always drops the card until a manual reboot).

## Fix

Use `opendir(SD_MOUNT_POINT)` instead. It succeeds only when the VFS path is registered and the volume mounted — which is exactly the case where `esp_vfs_fat_sdcard_format_cfg()` left `s_card` valid (its internal remount succeeded). When the remount failed the driver has already recycled `s_card` and `opendir()` fails, so `s_mounted` correctly stays false and no stale handle is exposed.

`<dirent.h>` is already included. 1 file, 2 code lines (+ comment).

## Testing

Code-review fix — the failure path (format fails while the card stays mounted) needs fault injection to hit directly. `opendir()` on the mount root is the same call the normal successful-format path relies on. Built with ESP-IDF v5.5.1 and flashed; normal Format SD behaves as before.

### DIFF
--- a/main/sd_storage.c
+++ b/main/sd_storage.c
@@ -359,14 +359,15 @@ esp_err_t sd_storage_format(void)
         if (ret != ESP_OK) {
             ESP_LOGE(TAG, "format: esp_vfs_fat_sdcard_format_cfg failed: %s",
                      esp_err_to_name(ret));
-            /* esp_vfs_fat_sdcard_format_cfg unmounts before formatting.  If
-             * f_mkfs failed, the volume may or may not have been remounted.
-             * Probe by trying a trivial VFS op; if it works, the card is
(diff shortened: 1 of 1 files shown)


ilyakruchinin:
Thank you @cwm12345 , merged!
Will be released in the next version shortly.

PR #187, opened 2026-09-02
fix(session_writer): queue a deferred export when EDF generation fails after a stop

### LINKED ISSUE #186: A session's EDF export is silently skipped when the uploader is mid-run at therapy stop — `edf_gen_generate` gives up after 120 s and nothing retries (main @ 9f54bf1)
Read on main @ 9f54bf1 (v1.4.1 + 3). I can't flash, so line refs and a suggested diff; the fix reuses the pending-export marker that already exists.

### What happens

`edf_gen.c:3704-3708`:

```c
bool have_lease = sd_storage_lease_acquire(SD_LEASE_EXPORT, 120000);
if (!have_lease) {
    ESP_LOGW(TAG, "export lease unavailable, skipping EDF generation");
    return ESP_ERR_TIMEOUT;
}
```

`sd_storage.c:280-300`: `SD_LEASE_EXPORT` and `SD_LEASE_UPLOAD` are the same recursive semaphore (`s_export_sem`).

`components/uploader/upload_sched.c:274` takes the UPLOAD lease (`uploader_lease_take` → `sd_storage_lease_acquire(SD_LEASE_UPLOAD)` via `net_provision.c:2059`) at the start of `run_backend` and only releases it at `:481`, after the **whole run** — every pending day, each with `day_begin` → groups → bundle → `day_end`. For SleepHQ `day_end` polls `shq_wait_import` up to 30 × 2 s = 60 s per day (`uploader_sleephq.c:537-553`), on top of transfer time and 30 s socket timeouts. `uploader.h:192-195` documents the lease as taken "for the duration of each day it uploads"; the code holds it for the run.

`session_writer.c:1557-1570` (post worker, after a stop):

```c
esp_err_t ret = edf_gen_generate(session_dir, session_id, start_epoch_ms, end_epoch_ms, drift_ms);
if (ret == ESP_OK) {
    ...
    uploader_on_export_complete(day_folder);
} else {
    ESP_LOGW(TAG, "post: EDF generation failed, not triggering upload");
}
```

No retry, and `pending_export_mark()` (`:1201`) is only called from boot-time crash recovery (`:2733`, `:2909`) for non-final manifests. The session's `.snt` files stay on the card, but nothing re-processes them except the manual "recreate EDFs" action (`net_provision.c:2287`). The night is missing from `DATALOG/` and from every upload backend.

### Concrete sequence

1. Wi-Fi/NAS/SleepHQ was unreachable for two nights; it's back. The 10-min scan starts a run for 2 pending days. Lease taken at `:274`.
2. User wakes up, stops therapy (or takes a mask-off break, restarts, stops again a few minutes later).
3. Post worker reaches `edf_gen_generate`, blocks 120 s on the lease.
4. The run is still in day 1's `shq_wait_import` (up to 60 s) → day 2 upload → day 2 `shq_wait_import` (up to 60 s). Two import waits alone can consume the whole 120 s before a byte of the second day is transferred.
5. `ESP_ERR_TIMEOUT`, "EDF generation failed, not triggering upload". Session never exported.

Only assumption: a run exceeds 120 s. With ≥ 2 backlog days on SleepHQ the import-wait alone makes that routine; a single large day over slow Wi-Fi does it too.

### Suggested fix (minimal, uses existing machinery)

`pending_export_service()` (`session_writer.c:1296`) already drains `SD_APP_DIR/pending_export/<day>.json` markers from the idle post worker and rebuilds the day, with attempt counting and stall handling. Marking the day on failure makes the export durable instead of lost:

```diff
--- a/main/session_writer.c
+++ b/main/session_writer.c
@@ -1557,14 +1557,20 @@
         esp_err_t ret = edf_gen_generate(session_dir, session_id,
                                         start_epoch_ms, end_epoch_ms, drift_ms);
+        char day_folder[32];
+        {
+            time_t t = (time_t)(start_epoch_ms / 1000);
+            struct tm tm;
+            localtime_r(&t, &tm);
+            if (tm.tm_hour < 12) { t -= 86400; localtime_r(&t, &tm); }
+            snprintf(day_folder, sizeof(day_folder), "%04d%02d%02d",
+                     tm.tm_year + 1900, tm.tm_mon + 1, tm.tm_mday);
+        }
         if (ret == ESP_OK) {
-            char day_folder[32];
-            time_t t = (time_t)(start_epoch_ms / 1000);
-            struct tm tm;
-            localtime_r(&t, &tm);
-            if (tm.tm_hour < 12) { t -= 86400; localtime_r(&t, &tm); }
-            snprintf(day_folder, sizeof(day_folder), "%04d%02d%02d",
-                     tm.tm_year + 1900, tm.tm_mon + 1, tm.tm_mday);
             uploader_on_export_complete(day_folder);
         } else {
-            ESP_LOGW(TAG, "post: EDF generation failed, not triggering upload");
+            ESP_LOGW(TAG, "post: EDF generation failed (%s), marking %s for deferred export",
+                     esp_err_to_name(ret), day_folder);
+            pending_export_mark(day_folder);
         }
```

Optionally also narrow the uploader's hold to match its own doc comment — take/give the lease around each day in the `:322` loop and around each oximetry put — so a stop during a long backlog run waits seconds instead of minutes. The marker alone already removes the data loss.

Plantucha:
Fixes #186.

> **Disclaimer: I don't have the hardware.** This change is not compiled or flashed — it is written from reading the code on `main @ 9f54bf1`. Please build and sanity-check it before merging; the change is confined to one block of `session_writer.c` and calls only functions already used in the same file.

## Problem

After a therapy stop the post worker calls `edf_gen_generate()`, which waits up to 120 s for `SD_LEASE_EXPORT`. That lease is the same recursive semaphore as `SD_LEASE_UPLOAD`, and `run_backend()` holds `SD_LEASE_UPLOAD` for the whole upload run (every pending day, including the SleepHQ import wait of up to 60 s per day). When a run is in progress at stop time, the export times out, the worker logs "EDF generation failed, not triggering upload", and nothing retries: the session's `.snt` files stay on the card but the night never appears in `DATALOG/` or on any backend until the user runs "Recreate EDFs" by hand. Details and a concrete sequence in #186.

## Fix

On export failure, leave a pending-export marker for the session's day via the existing `pending_export_mark()`. The idle post worker already drains these markers (`pending_export_service()`) with attempt counting and stall handling, so the day gets rebuilt once the lease is free instead of being lost. The `day_folder` computation moves ahead of the `if` so both branches can use it; the success path is unchanged.

`pending_export_mark()` is idempotent (an existing marker is left alone) and `esp_err_to_name()` is already used in this file. 1 file, +11/−3 lines.

## Testing

Not built or run — no ESP32 here. To reproduce the original problem: have ≥2 days pending for SleepHQ, let the 10-min scan start a run, then stop therapy while the run is in progress; the post worker should log the timeout and, with this change, `day <YYYYMMDD> queued for automatic export rebuild`, followed later by `pending export: rebuilding day …` once the uploader releases the lease.

Not addressed here (optional follow-up from #186): narrowing the uploader's lease hold to per-day, as `uploader.h:192-195` describes, so a stop during a long run waits seconds rather than minutes.

### DIFF
--- a/main/session_writer.c
+++ b/main/session_writer.c
@@ -1357,15 +1357,40 @@ static void pending_export_service(void)
         return;
     }
 
+    /* Probe the export lease without waiting.  This runs on the post-therapy
+     * worker, which also services therapy-stop jobs; blocking here for the
+     * rebuild's own 120 s lease timeout would stall those behind it.  If the
+     * uploader still holds the lease, leave the marker and try again on the
+     * next idle pass — the marker is durable, and this path deliberately does
+     * not count as an attempt, since nothing was attempted.
+     *
+     * The lease is released again immediately: this is a "busy right now?"
+     * question, not a hand-off.  edf_gen_rebuild_day() acquires it properly.
+     * If another task takes it in between, the rebuild falls back to its own
+     * timeout, which is exactly today's behaviour. */
+    if (!sd_storage_lease_acquire(SD_LEASE_EXPORT, 0)) {
+        ESP_LOGI(TAG, "pending export: day %s left pending — export lease busy",
+                 day);
+        return;
+    }
+    sd_storage_lease_release(SD_LEASE_EXPORT);
+
     ESP_LOGW(TAG, "pending export: rebuilding day %s (attempt %d)",
              day, attempts + 1);
     crash_diag_note_activity("rebuild_day");
 
     esp_err_t ret = edf_gen_rebuild_day(day);
     if (ret == ESP_OK) {
-        /* The day's files were replaced, so whatever the backends hold for it
-         * is stale.  Re-offer it, then retire the marker. */
-        uploader_on_day_invalidated(day);
+        /* Re-offer the day, then retire the marker.
+         *
+         * Deliberately NOT uploader_on_day_invalidated(): that forgets the
+         * whole day's upload index, so every session group from the night is
+         * re-offered as new.  A backend that opens an import batch per offer
+         * (SleepHQ) would then import a second time any earlier session from
+         * that night which had already uploaded cleanly.
+         * uploader_on_export_complete() reconciles instead — it marks only new
+         * or changed groups pending and leaves finished ones alone. */
+        uploader_on_export_complete(day);
         unlink(path);
         ESP_LOGW(TAG, "pending export: day %s rebuilt and queued for upload",
                  day);
@@ -1567,14 +1592,22 @@ static void sw_post_task(void *arg)
         }
         if (ret == ESP_OK) {
             uploader_on_export_complete(day_folder);
-        } else {
-            /* Most often the export lease was held by a long upload run
-             * (edf_gen_generate gives up after 120 s).  Leave a pending-export
-             * marker so the idle worker rebuilds the day later instead of the
-             * session silently never being exported. */
-            ESP_LOGW(TAG, "post: EDF generation failed (%s), marking day %s "
-                     "for deferred export", esp_err_to_name(ret), day_folder);
+        } else if (ret == ESP_ERR_TIMEOUT) {
+            /* The export lease was held by a long upload run (edf_gen_generate
+             * gives up after 120 s).  Nothing is wrong with the session itself,
+             * so leave a pending-export marker and let the idle worker rebuild
+             * the day rather than losing the night.
+             *
+             * Only this error defers.  A malformed session or an invalid
+             * timestamp fails identically on every retry, so marking those
+             * would burn the attempt budget and end in a stalled marker that
+             * needs attention, having achieved nothing. */
+            ESP_LOGW(TAG, "post: export lease busy, marking day %s "
+                     "for deferred export", day_folder);
             pending_export_mark(day_folder);
+        } else {
+            ESP_LOGW(TAG, "post: EDF generation failed (%s), not triggering upload",
+                     esp_err_to_name(ret));
         }
 
         UBaseType_t free_bytes = uxTaskGetStackHighWaterMark(NULL) * sizeof(StackType_t);
--- a/main/session_writer.c
+++ b/main/session_writer.c
@@ -1556,17 +1556,25 @@ static void sw_post_task(void *arg)
          * at the exact moment a session needs exporting. */
         esp_err_t ret = edf_gen_generate(session_dir, session_id,
                                         start_epoch_ms, end_epoch_ms, drift_ms);
-        if (ret == ESP_OK) {
-            char day_folder[32];
+        char day_folder[32];
+        {
             time_t t = (time_t)(start_epoch_ms / 1000);
             struct tm tm;
             localtime_r(&t, &tm);
             if (tm.tm_hour < 12) { t -= 86400; localtime_r(&t, &tm); }
             snprintf(day_folder, sizeof(day_folder), "%04d%02d%02d",
                      tm.tm_year + 1900, tm.tm_mon + 1, tm.tm_mday);
+        }
+        if (ret == ESP_OK) {
             uploader_on_export_complete(day_folder);
         } else {
-            ESP_LOGW(TAG, "post: EDF generation failed, not triggering upload");
+            /* Most often the export lease was held by a long upload run
+             * (edf_gen_generate gives up after 120 s).  Leave a pending-export
+             * marker so the idle worker rebuilds the day later instead of the
+             * session silently never being exported. */
+            ESP_LOGW(TAG, "post: EDF generation failed (%s), marking day %s "
+                     "for deferred export", esp_err_to_name(ret), day_folder);
+            pending_export_mark(day_folder);
         }
 
         UBaseType_t free_bytes = uxTaskGetStackHighWaterMark(NULL) * sizeof(StackType_t);


Plantucha:
I have read the CLA Document and I hereby sign the CLA

ilyakruchinin:
Hi @Plantucha,

Thank you very much for diagnosing #186 so thoroughly and for putting together this PR! The sequence you identified—where a therapy stop during an active upload pass causes `edf_gen_generate()` to time out on `SD_LEASE_EXPORT` and silently drop the session—is spot on, and your PR successfully eliminates that data-loss scenario.

Before merging, I'd like to ask for a few adjustments to make the recovery mechanism narrower and avoid potential side effects with backends like SleepHQ.

---

### Key Considerations

1. **Whole-Day Invalidation & SleepHQ Duplicate Imports**:
   - `pending_export_mark()` was originally written for boot-time crash recovery. When the idle worker services the marker via `pending_export_service()`, it rebuilds the day and calls:
     ```c
     uploader_on_day_invalidated(day);
     ```
   - In `upload_sched.c`, `EV_INVALIDATE` calls `upload_index_forget_day(day)`, which completely wipes the upload index for that day.
   - When the uploader next reconciles, it treats **every** session group from that night as new.
   - For SleepHQ, this creates a new import session (`POST /api/v1/teams/<team>/imports`) and re-uploads all files for the entire night. If a user had an earlier session before a midnight break that had already uploaded cleanly, that earlier session will be re-uploaded to SleepHQ in a second import batch.

2. **Idle Worker Blocking on Lease Contention**:
   - The idle worker services pending markers after 30 seconds of queue inactivity (`session_writer.c:1448`).
   - `edf_gen_rebuild_day()` then attempts to acquire `SD_LEASE_EXPORT` with a 120-second timeout (`edf_gen.c:4484`).
   - If the long upload pass that caused the initial contention is still running, the post-therapy worker task (`sw_post_task`) will block for up to another 120 seconds, during which it cannot service incoming therapy stop jobs from `s_post_q`.

3. **Error Scope**:
   - Marking the day on any non-`ESP_OK` return means unrecoverable errors (e.g. malformed data, invalid timestamps) will also queue deferred rebuilds and cycle up to 12 attempts before stalling.

---

### Suggested Changes

1. **Preserve Already-Uploaded Groups**:
   - Instead of wiping the day's index via `uploader_on_day_invalidated(day)`, notify the uploader via `uploader_on_export_complete(day)` once the deferred export completes.
   - `uploader_on_export_complete()` runs `reconcile_day_leased()`, which checks existing index entries and **only marks new or changed session groups as pending**, leaving already-uploaded sessions intact.

2. **Target Retryable Errors**:
   - Restrict the deferred mark specifically to lease contention or retryable errors:
     ```c
     if (ret == ESP_OK) {
         uploader_on_export_complete(day_folder);
     } else if (ret == ESP_ERR_TIMEOUT) {
         ESP_LOGW(TAG, "post: export lease busy, marking day %s for deferred export", day_folder);
         pending_export_mark(day_folder);
     } else {
         ESP_LOGW(TAG, "post: EDF generation failed (%s), not triggering upload", esp_err_to_name(ret));
     }
     ```

3. **Non-Blocking Lease Check in Idle Service (Optional / Recommended)**:
   - In `pending_export_service()`, perform a zero-timeout or short-timeout check on the lease before entering the rebuild. If the uploader is still busy, leave the marker alone and check again on the next idle cycle, rather than blocking the worker task for 120 seconds.

---

### Complementary Follow-Up
As you noted in #186, narrowing the uploader's lease hold in `upload_sched.c` to per-day (and releasing it while waiting on network polling like `shq_wait_import()`) will also be addressed as a follow-up. This will ensure that a stop during a backlog run rarely waits more than a few seconds in the first place.

Thanks again for the great investigation and contribution! Let me know what you think of this approach.

Plantucha:
All three adopted — pushed as a separate commit so your review stays anchored to the original.

The SleepHQ consequence is the one I had not thought through, and it is the
strongest of the three. My version fixed a lost night by risking a duplicated
one, which is a poor trade for anyone whose evening and morning sessions land
either side of midnight. `uploader_on_export_complete()` is clearly right.

**1. Reconcile instead of invalidate** — `pending_export_service()` now calls
`uploader_on_export_complete(day)`. Worth noting this also changes the
boot-time crash-recovery path, since it shares that call; I took that as
intended, because the same duplicate-import argument applies there. Say if you
would rather I scope it to the post-therapy case only.

**2. Defer only on `ESP_ERR_TIMEOUT`** — as you wrote it. I checked that this
is actually what lease contention returns before relying on it
(`edf_gen.c:3707`, `return ESP_ERR_TIMEOUT` after the 120 s acquire fails). If
it had returned `ESP_FAIL` the narrowed condition would have quietly turned
the whole PR into a no-op, so it seemed worth confirming rather than assuming.

**3. Non-blocking lease check** — taken, using a 0 ms acquire followed by an
immediate release, which `sd_storage.h:109` documents as the fail-fast form.
Two details I decided and would rather flag than bury:

- A deferral because the lease was busy **does not count as an attempt**.
  Nothing was attempted, and a long backlog could otherwise burn the 12-attempt
  budget without ever trying a rebuild, ending in a stalled marker.
- The probe is a "busy right now?" question, not a hand-off. If another task
  takes the lease between the probe and `edf_gen_rebuild_day()`, the rebuild
  falls back to its own 120 s timeout — i.e. exactly the current behaviour. It
  removes the common stall, not the race.

On the complementary follow-up: narrowing the uploader's hold in
`upload_sched.c`, and releasing it around `shq_wait_import()`, is the change
that makes this whole path rare rather than merely recoverable. Happy to leave
that with you, or to take a look if it would help.

One caveat, unchanged from the original PR: I have an AirSense 11 but no ESP32
running this firmware, so none of this has been built with ESP-IDF or run on a
device. It is reviewed code, not tested code — the lease semantics above are
read from the source rather than observed.

ilyakruchinin:
Thank you.
I'm working on a pretty big change at the moment addressing the issues you raised in discussion 191.
I will revisit/review this PR a bit later (once I do some testing of the "big change").

ilyakruchinin:
### NOTE TO SELF:
In sw_post_task(), day_folder is still derived using the pre-existing (tm.tm_hour < 12 ? t - 86400 : t) calculation rather than as11_time_noon_day(start_epoch_ms - drift_ms, day_folder, sizeof(day_folder)). PR #187 did not introduce this (it was already in main), so it does not block merging. After you merge PR #187, we can do a single-line cleanup to use as11_time_noon_day() if you want 100% uniformity with edf_gen.c.

@Plantucha - the above is for myself, no action is required from you.
**Your PR is in great shape, I will merge it** once I test the big change (just flagging some minor follow up due to the changes that I have pending).

Thank you for your input! Will be merged (a bit later, pending my other change validation).

ilyakruchinin:
Merged. Thank you @Plantucha

PR #192, opened 2026-09-02
fix(edf): round STR settings to the AS11 grid instead of truncating

Plantucha:
STR settings are written at 50 digits per unit, and therapy pressure moves in
0.2 cmH2O steps — so every settings pressure the AS11 stores is an exact
multiple of ten digits. 4.6 cmH2O is 230.

`(int16_t)(v->valuedouble * 50)` truncates toward zero. Doubles that are not
exactly representable land just below the integer — `4.6 * 50` is
`229.99999999999997` — so the cast yields **229**: one LSB low, and a value
off the grid the machine itself uses. It affects every pressure whose 0.2
step is not a clean binary fraction (4.6, 8.2, 9.2, 10.2, 16.4, …), and the
same truncation applies to the Ti fields, which use the same ×50 output
scale.

This replaces all 38 conversion sites with a rounding helper.

## Evidence

Checked against a real AirSense 11's SD card. Its `STR.edf` declares the
settings pressures as `phys[4,20]` against `dig[200,1000]` — exactly 50
digits per cmH2O — and every `S.C.` / `S.A.` / `S.AFH.` value stored in it is
a whole multiple of ten digits. The measured statistics in the same file
(`BlowPress.95` and friends) are *not* on that grid, which is what you would
expect: the grid is a property of settings, not of measurements.

So 229 is not merely one less than the ideal value; it is a value an AS11
would not write.

## How it was verified

I have host unit tests for `edf_gen.c` that I run locally. The STR assertion
— that every settings pressure is a multiple of ten digits — fails on `main`
and passes with this change:

```
main   : STR[6] 4.6 cmH2O stored as 229, which is not a multiple of 10
patched: XPASS
```

Those tests are not part of this PR; I mention them only to say the change
was executed rather than only read.

## Not verified

I do not have an ESP32 running this firmware, so this has not been built with
ESP-IDF or run on a device. The change is mechanical and identical at all 38
sites, but it does touch the VAuto/iVAPS settings, which my CPAP-only machine
never populates — I could not exercise those paths with real data.

`lrint()` needs `<math.h>`, which is added.

### DIFF
--- a/main/edf_gen.c
+++ b/main/edf_gen.c
@@ -33,11 +33,23 @@
 #include <time.h>
 #include <errno.h>
 #include <unistd.h>
+#include <math.h>
 
 #include "esp_log.h"
 #include "esp_timer.h"
 #include "esp_heap_caps.h"
 #include "cJSON.h"
+
+/* Settings are written at 50 digits per unit (see spec/0002-edf-export.md
+ * §4.3.4).  Round rather than truncate: therapy pressure moves in 0.2 cmH2O
+ * steps, so every settings pressure the AS11 writes is an exact multiple of
+ * ten digits -- 4.6 cmH2O is 230.  (int16_t)(4.6 * 50) yields 229, which is
+ * off that grid and one LSB low. */
+static inline int16_t settings_x50(double v)
+{
+    return (int16_t)lrint(v * 50.0);
+}
+
 
 static const char *TAG = "edf_gen";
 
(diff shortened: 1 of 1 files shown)


ilyakruchinin:
Merged!
Thank you, @Plantucha

ilyakruchinin:
Released in https://github.com/ilyakruchinin/SomnoTrace/releases/tag/v1.4.2

PR #197, opened 2026-09-03
fix(edf): CSL.edf must write "CSR Start" and "CSR End", not "CSR" for both

### LINKED ISSUE #190: CSL.edf labels both CSR edges "CSR"; the AS11 writes "CSR Start" and "CSR End"
`event_label_map_str()` maps every CSR edge to the single label `"CSR"`:

```c
/* main/edf_gen.c:3085-3090 */
if (csl_mode) {
    if (strcmp(ev_name, "CsrStart") == 0 || strcmp(ev_name, "CsrEnd") == 0 ||
        strcmp(ev_name, "CSRStart") == 0 || strcmp(ev_name, "CSREnd") == 0) {
        return "CSR";
    }
    return NULL;
}
```

The AS11 writes **two** labels. I checked my own AirSense 11's SD card (301
`CSL.edf` files): CSR episodes appear as a pair of zero-duration markers,

```
+<onset>\x150\x14CSR Start\x14\x00
+<onset>\x150\x14CSR End\x14\x00
```

alongside the `Recording starts` marker that every file has. So a
SomnoTrace-generated `CSL.edf` carries a label the machine never emits, and
the start/end distinction is gone — both annotations read `CSR`, and since
both are written with the same duration there is nothing left to tell them
apart or to reconstruct the episode span from.

There is a second-order effect in the same function. `generate_eve_edf()`
drops an event when the previous one has the same `onset_sec` **and** the
same label:

```c
if (dedup_count > 0 &&
    ev_list[dedup_count - 1].onset_sec == ev_list[i].onset_sec &&
    strcmp(ev_list[dedup_count - 1].label, ev_list[i].label) == 0) {
    continue;
}
```

That is correct for BLE retransmits of one event, but once both edges are
labelled `"CSR"` it also merges a genuine start/end pair that lands in the
same second — a short CSR episode becomes one annotation.

Suggested fix, returning the machine's own strings:

```c
if (csl_mode) {
    if (strcmp(ev_name, "CsrStart") == 0 || strcmp(ev_name, "CSRStart") == 0)
        return "CSR Start";
    if (strcmp(ev_name, "CsrEnd") == 0 || strcmp(ev_name, "CSREnd") == 0)
        return "CSR End";
    return NULL;
}
```

**One thing I could not check, and it may matter more than the above:** are
`CsrStart` / `CsrEnd` (or the `CSRStart` / `CSREnd` spellings) actually what
the AS11 sends under `TherapyEvents-RespiratoryEvents`? The four variants in
the current code suggest the name was not certain when it was written. I only
have the machine's SD card, not a BLE capture of its event stream, so I
cannot confirm any of them. If none of the four match, `CSL.edf` would
contain no CSR events at all rather than mislabelled ones — worth confirming
against a live capture before deciding how much this is worth.

I have not run this on an ESP32, so the above is from reading `edf_gen.c` at
9f54bf1 against real AS11 output — no on-device verification.

Plantucha:
Closes #190. You said you'd fix this yourself — if you already have, just
close this; I'd rather hand you the patch than have you redo it.

`event_label_map_str()` mapped every CSR edge to the single label `"CSR"`.
The AS11 writes two, verified across 301 real `CSL.edf` files on my own
AirSense 11's card, where CSR episodes appear as a pair of zero-duration
markers:

```
+<onset>\x150\x14CSR Start\x14\x00
+<onset>\x150\x14CSR End\x14\x00
```

So the generated `CSL.edf` carried a label the machine never emits, and the
start/end distinction was gone — both annotations read `CSR` with the same
duration, leaving nothing to pair them by or to reconstruct the span from.

There is a second-order effect in the same function that the label change also
fixes. `generate_eve_edf()` drops an event when the previous one has the same
`onset_sec` **and** the same label. That is right for BLE retransmits of a
single event, but once both edges carry the same label it also merges a
genuine start/end pair landing in the same second — a short CSR episode
became one annotation.

Your own trace of the origin is the best evidence of why this was never
caught: `spec/archive/edf-esp-vs-as11.md:143` says the CSR codes were
"extrapolated from the +1 pattern; no CSR events have been observed in test
sessions to confirm."

## What this does not settle

It does **not** establish that `CsrStart`/`CsrEnd` (or the `CSRStart`/`CSREnd`
spellings) are the names the AS11 actually sends under
`TherapyEvents-RespiratoryEvents`. I have no BLE capture of that stream. If
none of the four match, `CSL.edf` contains no CSR events at all rather than
mislabelled ones — a different and larger bug that this change does not fix
and cannot detect. Worth a look if you have a capture with CSR in it.

Not built with ESP-IDF or run on a device: I have an AirSense 11 but no ESP32
running this firmware, so this is read against real machine output rather than
observed on hardware.

### DIFF
--- a/main/edf_gen.c
+++ b/main/edf_gen.c
@@ -3095,10 +3095,16 @@ static const char *event_label_map_str(const char *ev_name, bool csl_mode)
 {
     if (!ev_name) return NULL;
     if (csl_mode) {
-        if (strcmp(ev_name, "CsrStart") == 0 || strcmp(ev_name, "CsrEnd") == 0 ||
-            strcmp(ev_name, "CSRStart") == 0 || strcmp(ev_name, "CSREnd") == 0) {
-            return "CSR";
-        }
+        /* The AS11 writes the two edges as SEPARATE labels — "CSR Start" and
+         * "CSR End", each a zero-duration marker — and a reader pairs them to
+         * recover the episode.  Returning one label for both loses that: the
+         * two annotations become indistinguishable, and the dedup below (same
+         * onset_sec + same label) then merges a genuine pair that lands in the
+         * same second. */
+        if (strcmp(ev_name, "CsrStart") == 0 || strcmp(ev_name, "CSRStart") == 0)
+            return "CSR Start";
+        if (strcmp(ev_name, "CsrEnd") == 0 || strcmp(ev_name, "CSREnd") == 0)
+            return "CSR End";
         return NULL;
     }
     if (strcmp(ev_name, "HypopneaEnd") == 0)          return "Hypopnea";


ilyakruchinin:
Hi @Plantucha,

Thank you for opening this PR! We actually implemented and verified this exact fix on main in commit 512fdc1 (with attribution to your diagnosis in #190).

In addition to mapping "CSR Start" and "CSR End", our implementation in main/edf_annotations.c also accounts for the backdateSeconds onset offset subtraction, enforces zero-duration boundary markers, and includes a host unit test in scripts/as11_events_test.c (Test 8).

Since edf_gen.c was recently modularized into edf_annotations.c and the fix is already live, I am going to close this PR. Thanks again for your great catch on #190!

PR #203, opened 2026-09-04
test(properties): bind the host property suite to the shipped code (#202)

### LINKED ISSUE #202: Property test suite exercises copies of the EDF code rather than the code, and passes only in AEST
I went over v2.0.0-pre2 against my AirSense 11 and its SD card. **Most of what I checked came back clean**, and I want to lead with that because it is the larger result:

- **The data dictionary matches the hardware.** I derived a signal table from 1488 files my machine wrote (2026-01-11 → 2026-09-02) and diffed it against `edf_data_dict.h`. Zero disagreements — every unit, physical range, digital range and samples-per-record — across BRP (2), PLD (9), SA2 (2), and all **77 STR labels my AutoSet writes**. `Crc16` is appended by `edf_header.c:217`, which accounts for the count differences.
- **The #200 DST fix is correct.** I ran `71e2e646`'s actual arithmetic (`floor_div`/`floor_mod`/`civil_from_days`, copied verbatim) through the harness that proved the original bug: 0 wrong in Europe/Prague, Europe/London, America/New_York and Australia/Sydney, with ordinary-night controls. `days -= 1` on a civil day number is the right shape.
- **`as11_time.c` is genuinely well tested.** I mutated it 6 ways — noon boundary ±1 s, `<` → `<=`, dropping the previous-day step, offset sign flip, the period-start noon anchor, plus a canary. **All 6 died.**

The rest of this is about `scripts/edf_properties_test.c`. **Nothing here affects a user today** — it is test coverage, not a live defect.

---

## 1 · The property suite exercises copies of the code, not the code

`edf_properties_test.c` includes only `esp_log.h` and `as11_time.h` from `main/`. It does not include `edf_data_dict.h`, `edf_summary.h`, `edf_waveform.h`, `edf_header.h`, or `edf_annotations.h`. Instead it defines its own `spool_to_edf`, `clamp_i16`, `snt_missing_for`, `is_channel_map_valid` and `format_recording_id`.

Those copies are **byte-identical to the originals today** — I diffed `spool_to_edf` against `edf_summary.c:448` line for line. That is exactly what makes it hard to notice: the suite is green, and it will stay green if the shipped versions change.

Two properties do not call anything at all:

```c
/* PROP_09 */
int brp_count = 2;
TEST_ASSERT(brp_count == 2, "PROP_09: BRP signal count is exactly 2");
```

```c
/* PROP_08 */
if (passthrough_true && stored == SNT_MISSING_V2) { out_val = -1; }
TEST_ASSERT(out_val == -1, "PROP_08: Passthrough signal outputs exact -1 sentinel");
```

PROP_09 asserts that 2 == 2; it never reads `EDF_BRP_SIGNAL_COUNT`. PROP_08 inlines the passthrough branch rather than calling the writer that implements it.

### The experiment

Three mutants against the shipped tree, run with `TZ=Australia/Sydney` (see §2 for why that matters). **Mutant C is the control — it is what makes the other two mean anything.**

| # | mutation | file | result |
|---|---|---|---|
| A | `spool_to_edf` rounding → `prod / den` (truncation) | `main/edf_summary.c` **shipped** | **37 passed, 0 failed — SURVIVED** |
| B | `EDF_BRP_SIGNAL_COUNT` 2 → 7 | `main/edf_data_dict.h` **shipped** | **37 passed, 0 failed — SURVIVED** |
| C | `if (tod < 43200) days -= 1;` → `days -= 0;` | `main/as11_time.c` *linked* | 34 passed, **3 failed — killed** |

Mutant A is worth pausing on: it reverts `spool_to_edf` to the truncating form — **the STR rounding bug from #192** — and the suite that ships alongside that fix does not notice. If that regression were reintroduced tomorrow, CI would stay green.

Reproduce:

```sh
gcc -O0 -o /tmp/pp scripts/edf_properties_test.c main/as11_time.c <cJSON.c> \
    -Imain -Iscripts/test_include -I<cjson dir> -lm
sed -i 's|int32_t rounded = (prod >= 0) ? (prod + half) / den : (prod - half) / den;|int32_t rounded = prod / den;|' main/edf_summary.c
# rebuild, rerun: still 37/37
```

### Suggested fix

Add the real translation units to the test build and delete the local copies, so the properties bind to shipped symbols:

```
main/edf_summary.c  main/edf_waveform.c  main/edf_header.c  main/edf_annotations.c
```

For PROP_09 specifically, read the macros rather than literals:

```c
TEST_ASSERT(EDF_BRP_SIGNAL_COUNT == 2, "...");
```

`as11_time_test.c` already does this correctly — it links `as11_time.c` and its mutants die. It is the model.

---

## 2 · The suite passes only in the author's timezone

There is no `setenv("TZ", …)` / `tzset()` anywhere in `edf_properties_test.c`. PROP_01's comment pins its expectations to AEST (*"2026-09-02 12:00:00 AEST (+10:00)"*), but `as11_time_noon_day()` falls back to `noon_day_esp_local()` when no offset is stored — and at that point in `main()` none is, because PROP_02 is what parses `+10:00` and it runs afterwards. So PROP_01 silently inherits the **host's** timezone.

| TZ | result |
|---|---|
| `Australia/Sydney` | **37 passed, 0 failed** |
| `UTC` | 36 passed, **1 failed** |
| `Europe/Prague` | 36 passed, **1 failed** |
| `America/New_York` | 34 passed, **3 failed** |

```
[FAIL] PROP_01: 12:00:00 PM maps to current day (D) (line 160)
[FAIL] PROP_01: Evening session belongs to current day (D) (line 165)
[FAIL] PROP_01: Post-midnight session belongs to previous day (D) (line 170)
```

The release notes describe this as a *"Host-Executable Test Harness … without requiring physical hardware"*. It also requires a specific host timezone, and it fails under UTC, which is what a CI runner would use.

### Suggested fix

Better than pinning `TZ`: set the AS11 offset explicitly, so the property tests the AS11 noon rule rather than the host's clock —

```c
as11_time_set_offset(36000);   /* +10:00, matching the comment */
```

`as11_time_set_offset` is already used by `as11_time_test.c`'s `reset_offset()`. Pinning `TZ` in `main()` would also work and is a one-liner, but leaves the property dependent on host configuration.

---

## Scope of what I checked

I audited where I hold ground truth: the signal dictionary (against my card) and the time code (against a libc harness). I have **not** reviewed `edf_summary.c` (1834 lines), `edf_waveform.c`, `edf_header.c`'s layout logic, the `.crc` removal fix, or the CSL/CSR annotation path.

That is not an aside — those are precisely the files §1 shows the property suite does not reach either. The gap and my blind spot are the same gap.

Happy to send a PR for either half if useful.

Plantucha:
Fixes #202. Both halves, verified by mutation rather than by the suite going green.

## Copies, not code

`edf_properties_test.c` included only `esp_log.h` and `as11_time.h` from `main/`, and defined its own `clamp_i16`, `snt_missing_for`, `is_channel_map_valid`, `spool_to_edf` and `format_recording_id`. Two properties called nothing at all — PROP_09 declared `int brp_count = 2` and asserted it equalled 2 without ever reading `EDF_BRP_SIGNAL_COUNT`; PROP_08 inlined the passthrough branch rather than exercising the writer.

🔴 **One copy had drifted, which is worse than duplication.** The test's `format_recording_id` called `gmtime_r` where the shipped `as11_time_format_recording_id` calls `localtime_r`, with the comment *"Test environment uses gmtime_r for UTC baseline"*. PROP_10 and PROP_16 were therefore green in **every** timezone while asserting behaviour the firmware does not have. A copy that matches is merely invisible; a copy that has drifted is green and wrong.

**Fix:** delete the copies, include the shipped headers. `clamp_i16` / `snt_missing_for` / `is_channel_map_valid` were already in `snt_format.h`, and `as11_time_format_recording_id` was already public — only `spool_to_edf` needed a home, and it moved from `edf_summary.c` into `edf_data_dict.h` beside the tables it converts for. Both headers are host-portable (`stdint`/`stdbool`/`limits` only), so no new stubs are needed.

PROP_09 now reads the shipped macros **and** checks each table's length against its own count macro, since a macro and its table can disagree.

## Timezone

PROP_01's expectations are pinned to AEST in its comments, but nothing set an offset: `as11_time_noon_day()` falls back to ESP-local time when none is stored, and PROP_02 — which parses `+10:00` — runs *after* it. So PROP_01 read the host clock.

It now calls `as11_time_set_offset(10 * 3600, "test")` explicitly, and `main()` pins `TZ=UTC` for the recording-id properties, which are legitimately local-time-dependent now that they use the real formatter.

## Result

| | before | after |
|---|---|---|
| `Australia/Sydney` | 37 / 0 | **40 / 0** |
| `UTC` | 36 / **1** | **40 / 0** |
| `Europe/Prague` | 36 / **1** | **40 / 0** |
| `America/New_York` | 34 / **3** | **40 / 0** |

## Verified by mutation

A suite that was already green is not evidence that it now tests more, so:

| | mutation | before | after |
|---|---|---|---|
| A | `spool_to_edf` rounding → truncation (the #192 bug) | SURVIVED | **KILLED** |
| B | `EDF_BRP_SIGNAL_COUNT` 2 → 7 | SURVIVED | **KILLED** |
| C | noon-day previous-day step — **control** | killed | killed |

C is the control: killed both before and after, so A and B flipping is the change itself and not a difference in how the suite was run.

`as11_time_test` and `as11_events_test` still build and pass. Builds clean under `-Wall -Wextra`.

⚠️ One thing I could not do from here: I have no ESP-IDF toolchain, so I verified `edf_summary.c` only by host syntax parse after moving `spool_to_edf` out of it. A firmware build is worth a glance before merge.

### DIFF
--- a/main/edf_data_dict.h
+++ b/main/edf_data_dict.h
@@ -25,10 +25,32 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <limits.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+/* ── Spool → EDF fixed-point conversion ──────────────────────────────── */
+
+/* Convert a Summary-spool fixed-point integer to its EDF digital value.
+ * `raw == -1` is the missing sentinel and passes through untouched.
+ * Rounding is SYMMETRIC (half away from zero) — truncating here silently
+ * biased every STR settings value low, which is what issue #192 reported.
+ *
+ * Lives in this header rather than in edf_summary.c so the host property
+ * tests bind to the same definition the firmware uses. A test carrying its
+ * own copy stays green when this one changes. */
+static inline int16_t spool_to_edf(int16_t raw, int num, int den)
+{
+    if (raw == -1 || den <= 0) return -1;
+    int32_t prod = (int32_t)raw * num;
+    int32_t half = den / 2;
+    int32_t rounded = (prod >= 0) ? (prod + half) / den : (prod - half) / den;
+    if (rounded > INT16_MAX) return INT16_MAX;
+    if (rounded < INT16_MIN) return INT16_MIN;
+    return (int16_t)rounded;
+}
 
 /* ── Signal definition structure ─────────────────────────────────────── */
 
--- a/main/edf_summary.c
+++ b/main/edf_summary.c
@@ -445,17 +445,6 @@ static int profile_name_to_mop(const char *name)
  *   Indices (AHI etc)    logical_scale = 10    → num=1, den=10
  *   Duration/enums/SAU   logical_scale = 1     → no conversion needed
  */
-static inline int16_t spool_to_edf(int16_t raw, int num, int den)
-{
-    if (raw == -1 || den <= 0) return -1;
-    int32_t prod = (int32_t)raw * num;
-    int32_t half = den / 2;
-    int32_t rounded = (prod >= 0) ? (prod + half) / den : (prod - half) / den;
-    if (rounded > INT16_MAX) return INT16_MAX;
-    if (rounded < INT16_MIN) return INT16_MIN;
-    return (int16_t)rounded;
-}
-
 /* Build STR data values [4-132] from a summary context and settings JSON.
  * str_values must be pre-filled with 0xFF (sentinel for "no data").
  *
--- a/scripts/edf_properties_test.c
+++ b/scripts/edf_properties_test.c
@@ -37,64 +37,20 @@
 
 #include "cJSON.h"
 #include "esp_log.h"
+#include "snt_format.h"      /* clamp_i16, snt_missing_for, is_channel_map_valid */
+#include "edf_data_dict.h"   /* signal counts, spool_to_edf */
 #include "as11_time.h"
 
 /* ════════════════════════════════════════════════════════════════════
- *  Replicated Production Helpers Under Test (main/edf_gen.c)
+ *  Production code under test — included, never replicated.
+ *
+ *  These properties used to assert against private copies of clamp_i16,
+ *  snt_missing_for, is_channel_map_valid, spool_to_edf and the recording-id
+ *  formatter. A copy stays green when the shipped version changes, so the
+ *  suite could not see a regression in the code it names. They are included
+ *  from snt_format.h / edf_data_dict.h / as11_time.h now.
  * ════════════════════════════════════════════════════════════════════ */
 
-#define SNT_MISSING_V1  -1
-#define SNT_MISSING_V2  INT16_MIN
-
-static inline int16_t snt_missing_for(uint8_t version)
-{
-    return (version >= 2) ? SNT_MISSING_V2 : SNT_MISSING_V1;
-}
-
-static inline int16_t clamp_i16(int val, int16_t min_val, int16_t max_val)
-{
-    if (val > max_val) return max_val;
-    if (val < min_val) return min_val;
-    return (int16_t)val;
-}
-
-static inline bool is_channel_map_valid(const int *map, int n_signals, int snt_channels)
-{
-    if (!map) return (snt_channels == n_signals);
-    for (int i = 0; i < n_signals; i++) {
-        if (map[i] < 0 || map[i] >= snt_channels) return false;
-    }
-    return true;
-}
-
-static inline int16_t spool_to_edf(int16_t raw, int num, int den)
-{
-    if (raw == -1 || den <= 0) return -1;
-    int32_t prod = (int32_t)raw * num;
-    int32_t half = den / 2;
-    int32_t rounded = (prod >= 0) ? (prod + half) / den : (prod - half) / den;
-    if (rounded > INT16_MAX) return INT16_MAX;
-    if (rounded < INT16_MIN) return INT16_MIN;
-    return (int16_t)rounded;
-}
-
-static void format_recording_id(char *out, size_t out_len,
-                                int64_t epoch_ms,
-                                const char *srn, const char *mid, const char *vid)
-{
-    static const char *month_names[] = {
-        "JAN", "FEB", "MAR", "APR", "MAY", "JUN",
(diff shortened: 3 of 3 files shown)


Plantucha:
One offer, deliberately kept out of this PR so it doesn't cost review time.

The mutation check that produced the before/after table is a **scratch script** — hardcoded paths, a hand-picked mutant list, and cJSON found by a literal path on my box. It proved the fix but it is not something you'd want in the tree as-is.

If a permanent version would be useful I'm happy to write one properly. Shape:

- mutate a named shipped file, rebuild the host tests, report `SURVIVED` / `KILLED` per mutant
- **baseline** — the unmutated tree must pass first, with nothing skipped; if it doesn't, the run reports nothing
- **canary** — one mutant no working suite could possibly miss. If the canary survives, the harness is not running what it thinks it is and the whole run is **VOID** rather than reported as "0 survivors"
- non-zero exit on any survivor, so it can gate

The reason I think it's worth having rather than a one-off: **this fix can regress quietly.** The natural way to add PROP_17 is to reach for a helper, and copying one in is easier than wiring a header. Nothing in CI would object — the suite would be green, exactly as it was before this PR. Mutation is the only check that distinguishes "the tests pass" from "the tests would notice".

Happy to open it as a separate PR, or leave it entirely — your call, and no need to decide as part of merging this one.

For reference, the mutants used here were:

| | mutation | target |
|---|---|---|
| A | `spool_to_edf` rounding → `prod / den` | shipped `spool_to_edf` |
| B | `EDF_BRP_SIGNAL_COUNT` 2 → 7 | shipped macro |
| C | `if (tod < 43200) days -= 1;` → `days -= 0;` | `as11_time.c`, as the control |

C is the one that matters methodologically: it was killed both before and after, which is what makes A and B flipping attributable to this change rather than to a difference in how I ran the suite.

ilyakruchinin:
Merged!
Thank you @Plantucha !

PR #204, opened 2026-09-05
test(as11_time): cover civil-date arithmetic at the February and 2100 boundaries

Plantucha:
`days_from_civil()` shifts the year to start on March 1, and the shift fires only for January and February. Every existing scenario reaches it with a March-or-later date, so inverting either February branch (`m <= 2` → `m < 2`, `m > 2` → `m >= 2`) passed the whole host suite. The parser that sees arbitrary dates — `as11_time_parse_iso8601_ms()`, which handles every AS11 event timestamp — had no host test at all.

Scenario 9 adds 11 assertions: ISO event timestamps on the leap day, on both sides of the Feb 28 / Mar 1 boundary in a leap and a non-leap year, a malformed input, the STR day number for the leap day and the day after, and the 2100-03-01 century boundary in both directions (parse and noon-day label), which is the first date the `doe/36524` term in `civil_from_days()` is not inert.

Expected values come from an independent calendar (Python `datetime`), not from the functions under test. The firmware arithmetic agrees on all eleven; **this changes no shipped code**.

### Verification
- `scripts/as11_time_test.c` builds and passes on the host with the new scenario (all existing scenarios unchanged).
- Under mutation (`scripts/mutate_host.py` on the `tools/host-mutation-harness` branch) the three surviving mutants at `as11_time.c:57`, `:60` and `:70` are now killed.

Test-only, independent of #203; based directly on `v2.0.0`.

### DIFF
--- a/scripts/as11_time_test.c
+++ b/scripts/as11_time_test.c
@@ -192,6 +192,42 @@ int main(void)
     check_period_end("Feb 28 23:00 (leap year rollover)", 2024, 2, 28, 23, 0, "2024-02-29 12:00");
     check_period_end("Oct 24 23:00 (evening before fall-back)", 2026, 10, 24, 23, 0, "2026-10-25 12:00");
 
+    printf("\n=== Scenario 9: Civil-Date Arithmetic at the February Boundary ===\n");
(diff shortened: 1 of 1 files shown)


ilyakruchinin:
Merged!
Thank you @Plantucha !

PR #205, opened 2026-09-05
test(properties): PROP_04 covers a non-unity numerator in spool_to_edf

Plantucha:
> **Stacked on #203** — the diff will show #203's changes until it merges; this PR's own change is the single commit `1b16fae` (10 lines in `scripts/edf_properties_test.c`). It is stacked because the assertion has to bind to the shipped `spool_to_edf()`; on `main` it would test the private copy #202 is about.

Every PROP_04 case passed `num == 1`, which cannot distinguish `raw * num` from `raw / num`. The firmware does not: the three Flow channels pass `(5, 1)` and MinVent/TgtVent pass `(2, 25)` — `edf_summary.c:756-758`, `:787-789`, `:798-800`. Under the multiply→divide mutation Flow.95 is written 25× too small and the property suite stays green.

Five assertions: the Flow scale in both signs, and the `(2, 25)` scale on both sides of a rounding boundary and negative. Expected values are arithmetic, not read back from the function. **No shipped code changes.**

### Verification
- `edf_properties_test`: 45/45 pass on host (40 existing + 5 new).
- Under mutation (`scripts/mutate_host.py`, `tools/host-mutation-harness`) the survivor at `edf_data_dict.h:47` (`*` → `/`) is now killed.

### DIFF
--- a/main/edf_data_dict.h
+++ b/main/edf_data_dict.h
@@ -25,10 +25,32 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <limits.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+/* ── Spool → EDF fixed-point conversion ──────────────────────────────── */
+
+/* Convert a Summary-spool fixed-point integer to its EDF digital value.
+ * `raw == -1` is the missing sentinel and passes through untouched.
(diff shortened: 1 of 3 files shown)


ilyakruchinin:
Merged!
Thank you @Plantucha !

PR #206, opened 2026-09-05
test: assert the boundaries the mutation harness showed were never asserted

Plantucha:
> **Stacked on #204** — this adds Scenarios 10 and 11 after #204's Scenario 9 in `scripts/as11_time_test.c`; the diff shows #204's commit until it merges. This PR's own change is the single commit `b564500` (97 lines across two test files).

Running the host mutation harness over `as11_time.c` and `oximetry_vld3.c` (on v2.0.0 with #203–#205 applied) left 17 and 7 survivors respectively — mutants the suites executed but never noticed. 18 of them are test gaps and this PR closes them; the other 6 are equivalent mutants (a boundary where the original and the mutant agree at the only value that separates them), recorded with their arguments in the harness's `mutate_equivalent.txt`. **No shipped code changes**; every new expectation passes against the unmodified firmware.

### `as11_time_test.c` — Scenario 10, settings offset parser and `as11_time_set_offset()`
Every earlier settings case was a whole hour in the interior of [-12:00, +14:00], so `hh*3600 + mm*60` was indistinguishable from `hh*3600 - mm*60`, hour 0 was never parsed, neither bound was shown inclusive, and the `'+'` branch was only ever taken with a sign present. Adds half-hour offsets, `+00:00`, both bounds and one step past each, and an unsigned value.

### `as11_time_test.c` — Scenario 11, noon-stamp derivation
`PeriodStart == 0` (the `<= 0` guard), the 5-minute tolerance from both sides and one second past it, and `00:00Z` — local noon at +12:00 and −12:00 alike — resolved against anchors −10:00, +10:00 and 00:00: −12:00 (the lower bound itself), +12:00, and +12:00 (a tie keeps the raw candidate). The anchor is set through the settings parser so the result does not depend on the host TZ.

### `vld3_decoder_test.c`
The 2024-02-29 header was parsed but `datetime_valid` never asserted, so the leap-year rule, `month == 2`, and `day <= max_day` could all be inverted unnoticed. `asleep_seconds` and `duration_seconds` were decoded and never read. The year/hour/minute/second bounds had no case on either side. The NULL guards were untested.

### Verification
| file | before | after |
|---|---|---|
| `as11_time.c` | killed 44, unasserted 17 | killed 55, unasserted **0**, equivalent 8 |
| `oximetry_vld3.c` | killed 13, unasserted 7 | killed 20, unasserted **0** |

Both suites pass on host. Harness: `scripts/mutate_host.py` on the `tools/host-mutation-harness` branch (offered on #203).

### One observation, not changed here
`as11_time_parse_iso8601_ms()` (`as11_time.c:320-330`) retries with a second `sscanf` format when the first returns fewer than 6 fields — but the two formats are identical through the sixth conversion, so the retry can never succeed where the first failed. It's harmless dead code; the two mutants on those lines are recorded equivalent for that reason. Flagging in case you want to drop the retry during the rewrite.

### DIFF
--- a/scripts/as11_time_test.c
+++ b/scripts/as11_time_test.c
@@ -77,6 +77,19 @@ static void check_period_end(const char *what, int y, int m, int d, int h, int m
     expect_str(what, buf, want_dt);
 }
 
+/* Feed one TimeZoneOffset string through the settings parser, in the FlowGenerator shape
+ * Scenario 4 builds by hand. A successful parse also makes the value the anchor for every
+ * later noon-stamp derivation, which is what Scenario 11 relies on. */
+static bool parse_tz_setting(const char *value, as11_offset_t *out)
+{
+    cJSON n_off = { .child = NULL, .next = NULL, .string = "TimeZoneOffset", .valuestring = (char *)value, .type = cJSON_String };
+    cJSON n_tz = { .child = &n_off, .next = NULL, .string = "TimeZoneFeature", .valuestring = NULL, .type = 0 };
+    cJSON n_fp = { .child = &n_tz, .next = NULL, .string = "FeatureProfiles", .valuestring = NULL, .type = 0 };
+    cJSON n_sp = { .child = &n_fp, .next = NULL, .string = "SettingProfiles", .valuestring = NULL, .type = 0 };
+    cJSON n_fg = { .child = &n_sp, .next = NULL, .string = "FlowGenerator", .valuestring = NULL, .type = 0 };
+    return as11_time_offset_from_settings(&n_fg, out);
+}
+
 int main(void)
 {
     printf("\n=== Scenario 1: Issue #75 Vector (AS11 -05:00 vs ESP MST7MDT) ===\n");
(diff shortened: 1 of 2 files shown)


Plantucha:
One note on what this PR does and doesn't protect.

These assertions hold the line at v2.0.0, but nothing runs them: the only workflow in `.github/` is the CLA check, and the four host suites (`as11_time_test`, `as11_events_test`, `edf_properties_test`, `vld3_decoder_test`) are built by hand. A future commit can break any of them, or add a branch with no assertion behind it, and the first anyone hears of it is the next time someone runs gcc.

Two steps, if you want them — both are small and I'm happy to send either as a PR:

1. A workflow that builds and runs the four host suites on every PR. Seconds of runtime; it needs `cJSON.c` on the runner (a checkout of DaveGamble/cJSON at a pinned tag, or the copy from the IDF component registry).

2. On top of that, the mutation harness from #203 as a gate: `mutate_host.py --all` exits non-zero on any unasserted survivor, and with the equivalents recorded in `mutate_equivalent.txt` both `as11_time.c` and `oximetry_vld3.c` are at zero today, so the gate would go green on this branch and fail the first PR that adds unasserted logic to either file. A full run is about 30 s on a desktop (gcc -O0, `--coverage` for reach, every mutant rebuilt), so it fits inside the PR check rather than a nightly.

Step 1 is worth doing regardless of 2. If the rewrite you mentioned in #187 is close, it may make more sense to land these after it, against the new file layout — your call.

ilyakruchinin:
Merged!
Thank you @Plantucha !

Issue #200, opened 2026-09-04
Noon-day label is off by one day after a DST spring-forward (t -= 86400, 6 sites)


Plantucha:
The noon-day rule ("a session before local noon belongs to the previous day") is implemented in six places as epoch arithmetic:

```c
localtime_r(&t, &tm);
if (tm.tm_hour < 12) {
    t -= 86400;
    localtime_r(&t, &tm);
}
```

A local day is not always 86400 seconds. On the day a spring-forward happens it is 23 hours, so subtracting 86400 seconds from an instant shortly after local midnight lands on the day *before* the intended one. The resulting day label is off by one, and the session is filed — and uploaded — under the wrong day folder.

## Where

| Site | Rule | Shape |
|---|---|---|
| `main/session_writer.c:355` | `<12 → t -= 86400` | helper `noon_day_folder_local()` |
| `main/session_writer.c:1563` | `<12 → t -= 86400` | open-coded copy of that helper, same file |
| `main/net_provision.c:2257` | `<12 → t -= 86400` | inline |
| `main/post_therapy.c:474` | `<12 → t -= 86400` | inline, then pinned to noon via `mktime` |
| `main/post_therapy.c:626` | `>=12 → t += 86400` | `noon_period_end()` |
| `main/oximetry_canonical.c:230` | `<12 → t -= 86400` | `day_for_epoch()` |

`main/as11_time.c:218` implements the same rule against the AS11 clock rather than libc and is not affected by this.

## Reproduction

Self-contained, no ESP-IDF needed — `noon_day_folder_local` below is copied verbatim from `session_writer.c:355`:

```c
#include <stdio.h>
#include <string.h>
#include <time.h>

static void noon_day_folder_local(time_t t, char *out, size_t out_len)
{
    struct tm tm;
    localtime_r(&t, &tm);
    if (tm.tm_hour < 12) {
        t -= 86400;
        localtime_r(&t, &tm);
    }
    snprintf(out, out_len, "%04d%02d%02d",
             (tm.tm_year + 1900) % 10000, (tm.tm_mon + 1) % 100, tm.tm_mday % 100);
}

int main(void)
{
    struct tm tm; memset(&tm, 0, sizeof tm);
    tm.tm_year = 2026 - 1900; tm.tm_mon = 2; tm.tm_mday = 30;   /* 2026-03-30 */
    tm.tm_hour = 0; tm.tm_min = 30; tm.tm_isdst = -1;           /* 00:30 local */
[...]

PR #207, opened 2026-09-05
test: host tests for the EDF converter, with a mutation check that has controls

Plantucha:
Host-side (plain gcc, no ESP-IDF) tests for the EDF converter, with a mutation check that tells you whether the suite is actually measuring anything.

Stacked on #203 (its commit `f047f81` is the first one here; merge #203 and it drops out). The property suite needs its pinned offset/TZ to be green off a Melbourne host, and `mutants.py` refuses to report a number on a red baseline.

## What is in it

| file | what |
|---|---|
| `scripts/run_host_tests.sh` | builds and runs every `scripts/*_test.c`; refuses to run if a `*_test.c` exists it does not know about |
| `scripts/test_include/` | one-file stand-ins for `esp_log.h`, `esp_err.h`, `esp_timer.h`, `esp_heap_caps.h` |
| `scripts/edf_gen_test.c` | `#include`s `edf_header/waveform/annotations/summary/gen.c` and runs the real converter end to end: SNT files in, EDF bytes out. 16 tests |
| `scripts/mutants.py` | plants 15 one-line bugs into a copy of `main/` and expects the suite to fail on each |
| `scripts/mutants_probe.py` | optional; asks a local Ollama model for more one-line bugs and measures them the same way. Off by default, nothing depends on it |
| `scripts/HOST_TESTS.md` | how to run it, and the rules the tests follow (test the real code, no constants copied from `main/`, differential oracles where two paths must agree) |

## What the tests check

SNT v1 and v2 export identical EDF; the missing-data sentinel survives as `-1`; a torn `.snt` tail drops the partial record instead of promising it in the header; overflow saturates at the declared digital maximum; a PLD file shorter than its channel map is refused; MaskOff ends the export; the `_ZLE` gate applies the AS11→NTP drift; `as11_time_noon_day()` puts 11:59:59 on D-1 and 12:00:00 on D; STR pressure settings are exact cmH2O × 50; `spool_to_edf` keeps the `-1` sentinel.

## What `mutants.py` guards against

Two controls run before any mutant, and a failed control makes the run **VOID** with no kill count:
- **baseline** — the unmutated tree must pass with nothing skipped;
- **canary** — a mutant no working suite can miss (every SNT file rejected).

A kill needs positive evidence (the runner's own summary line naming the failing test); a mutant that does not compile is reported as malformed, not as killed; a mutant whose anchor no longer matches is `NOT APPLIED`, not silently skipped. Three survivors are listed as equivalent with the argument for each, and are still run — one the suite kills is reported as `REFUTED`.

## Result on v2.0.0 (`45a1309`) + #203

```
baseline   PASS          5 run, all passed
canary     killed        killed by edf_gen_test
15/15 mutants killed (canary passed, baseline green); 3 classified equivalent
```

## History

This suite was written against pre-2.0 `edf_gen.c` and carried two expected failures for bugs it found there (#192, #200). Both are fixed in v2.0.0, so it now has no XFAILs. The last commit ports it to the module split from 3623ea3 and the removal of `noon_day_folder()` in 71e2e64.

## Not in this PR

- No CI workflow. `HOST_TESTS.md` shows the two lines it would take; that is your call.
- `as11_events_test` still tests a copy of the parser rather than the shipped one (pre-existing).


ilyakruchinin:
Merged!
Thank you @Plantucha !

PR #208, opened 2026-09-05
fix(as11_ble): check the malloc fallback before using the buffer

Plantucha:
## What

`main/as11_ble.c` uses this pattern six times:

```c
char *rpc = heap_caps_malloc(N, MALLOC_CAP_SPIRAM);
if (!rpc) rpc = malloc(N);
```

Two of them check the fallback (`:1831` returns `ESP_ERR_NO_MEM`, `:1885` logs and returns). The other four use the buffer straight away, so if both allocations fail `snprintf` writes through NULL and the device resets from inside the BLE task:

| site (v2.0.0) | task | buffer | now |
|---|---|---|---|
| `pair_task`, StartKeyExchange | `json`, 800 B | `set_error(...)`, terminate the connection, delete the task |
| `reconnect_task`, Get (test) | `rpc`, 512 B | log, skip the step |
| `reconnect_task`, SubscribeEvent | `rpc`, 700 B | log, skip the step |
| `reconnect_task`, StartStream | `rpc`, 600 B | log, skip the step |

For `reconnect_task` the skip mirrors what already happens when `as11_ble_send_and_wait` fails at those steps (the reconnect carries on and the next connection cycle retries). For `pair_task` there is no partial-continue that makes sense, so it fails the pairing the same way the other `set_error` paths do.

Both allocations failing is only likely on a fragmented heap after long uptime, which is exactly the state the reconnect path runs in.

## How found

`cppcheck 2.19 --enable=warning,performance,portability --inconclusive` over `main/` reports these four as `nullPointerOutOfMemory`. They were the only substantive findings in the tree. `gcc -fanalyzer` over the host-buildable files (`edf_*.c`, `as11_time.c`, `oximetry_vld3.c`) reports nothing.

## Verification

- Clean build on `espressif/idf:v5.5.1` for esp32s3: `Project build complete`, 0 compiler warnings in `main/`.
- Not flashed: I don't have the hardware, and the OOM path can't be exercised without it. The change is four early-outs on the failure branch; the success path is untouched.

+21 lines, one file.

### DIFF
--- a/main/as11_ble.c
+++ b/main/as11_ble.c
@@ -1635,6 +1635,12 @@ static void pair_task(void *arg)
     bytes_to_hex(s_srp.A_pad, SRP_PAD, A_hex);
     char *json = heap_caps_malloc(800, MALLOC_CAP_SPIRAM);
     if (!json) json = malloc(800);
+    if (!json) {
+        set_error("StartKeyExchange: out of memory");
+        ble_gap_terminate(s_conn_handle, BLE_ERR_REM_USER_CONN_TERM);
+        vTaskDelete(NULL);
+        return;
+    }
     snprintf(json, 800,
              "{\"id\":1,\"jsonrpc\":\"2.0\",\"method\":\"StartKeyExchange\","
              "\"params\":{\"clientPk\":\"%s\"}}", A_hex);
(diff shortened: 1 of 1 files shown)


ilyakruchinin:
Merged!
Thank you @Plantucha !

PR #210, opened 2026-09-05
test(as11_time): cover the three header formatters

Plantucha:
`as11_time_format_edf_datetime()`, `_recording_id()` and `_session_prefix()` write the EDF startdate/starttime fields, the recording-id line and the session directory prefix, and no host test reached them: the edf suites build their headers through their own fixtures. This adds Scenario 12 to `scripts/as11_time_test.c` (+57 lines, one file, no production change).

What it pins:

- All three go through `localtime_r()`, so the scenario sets `TZ=EST5` (no DST rule, so the result does not depend on where in the year the date falls) and the expected strings were computed independently with Python `datetime` in that zone. In UTC a `gmtime_r()` swap would pass unnoticed; here it fails.
- Aug 7 2026 21:38:33.027Z (Scenario 7's evening session), Dec 31 (the last entry of the month-name table), Jan 5 2009 (two-digit year keeps its leading zero), and 2100-03-01T00:00Z — which the −5 h zone moves back to `28.02.00`, so date and year both come from the local conversion.
- NULL identifiers print as empty `SRN= MID= VID=`, not `(null)`; a NULL `time_out` still writes the date; `time_len` 0 leaves the buffer untouched.

Mutation spot-check on the formatters: gmtime swap, month index off by one, prefix month not +1, NULL srn printing `(null)`, and a canary (prefix separator) all fail the suite. Three mutants survive and are equivalent: dropping the `date_len > 0` / `time_len > 0` guards (snprintf with size 0 writes nothing), and `tm_year % 100` without the `+100` wrap (identical for any year ≥ 1900).

`as11_time.c` line coverage under the suite goes to 96% (176 lines); what is left unexecuted is the no-offset fallbacks in `noon_day_esp_local()`.

Host suite: `scripts/run_host_tests.sh` 5 run, 0 failed.

### DIFF
--- a/scripts/as11_time_test.c
+++ b/scripts/as11_time_test.c
@@ -294,6 +294,63 @@ int main(void)
                as11_time_offset_from_period_start(PS_MIDNIGHT_Z, &off) ? off : -999999, 43200);
     reset_offset();
 
+    printf("\n=== Scenario 12: Header Formatters ===\n");
+    /* The three formatters that write EDF header fields and the session prefix are
+     * reached by no host test: the edf suites build headers through their own
+     * fixtures. All three go through localtime_r(), so a fixed non-UTC zone is
+     * pinned and the expected strings come from Python datetime in that zone; in
+     * UTC a gmtime_r() swap would pass unnoticed. EST5 has no DST rule, so the
+     * result does not depend on the date's position in the year. */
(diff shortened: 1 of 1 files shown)


ilyakruchinin:

PR #211, opened 2026-09-05
Test connection buttons for SMB and SleepHQ (#123)

### LINKED ISSUE #123: [FR] - test button for SMB/SHQ

Plantucha:
Closes #123. Adds a "Test connection" button to the SMB and SleepHQ cards in the portal, backed by two endpoints that probe the settings **saved in NVS** without starting an upload.

### Endpoints

`POST /api/uploads/test-smb` and `POST /api/uploads/test-sleephq`, no body. Response is always `{"ok": bool, "message": "..."}`:

- `200` — the probe ran; `ok` says whether it succeeded, `message` says why not (or what it reached).
- `409` — an upload run is active, or the uploader has not finished init.
- `404` — unknown backend.

The portal button refuses with "Save settings first, then test" while the card has unsaved inputs, so the probe never tests something other than what the uploader will use.

### What each probe does

- **SMB**: `smb2_connect_share` on a private context (`smb2_set_timeout(10)`), then `smb2_stat` of the saved remote path. A missing folder is reported as a failure on purpose: the uploader's `smb2_mkdir` calls never create parents, so that configuration would fail on the first upload.
- **SleepHQ**: TLS connect (10 s) and one `/oauth/token` request; the token is discarded, no import is opened. `shq_authenticate()` short-circuited on a cached token, which would have made a "test" a no-op, so it is now `shq_request_token()` plus the cache wrapper. The upload path's behaviour is unchanged (same `ESP_FAIL` cases; a 4xx returns before the JSON parse instead of failing at "no access_token").

Wiring: an optional `bool (*test)(char *msg, size_t)` on `upload_backend_t`; `uploader_test_connection()` does the registry lookup and the busy check (`upload_sched_uploading()`). `spec/0011-web-api-endpoints.md` gets its first `/api/uploads/*` section (4.7) — the existing upload endpoints were not documented there before.

### Decisions you may want to change

1. The probe ignores the enable toggle, so a backend can be tested before it is enabled.
2. The busy check is check-then-run; a run starting inside the ~10 s probe window is not excluded. Closing that needs a probe flag the scheduler honours, which seemed too invasive for this.
3. The probe blocks the httpd worker for up to ~10 s (status polling stalls meanwhile) — same as `/api/alert/test` today.
4. The 409 body uses `{ok,message}` rather than `send_busy()`'s `{ok,error}`, to keep one shape per endpoint.
5. Therapy-active is not a guard: the code only defers periodic scans during recording and still runs event uploads, so the probe is not blocked either.

### Verification and its limits

Full `espressif/idf:v5.5.1` esp32s3 build: `Project build complete`, 0 warnings in the touched files, total warning count unchanged (6 → 6).

Not verified, no hardware here: not flashed; never run against a real SMB share or SleepHQ account; httpd worker stack headroom for a TLS handshake (the uploader does the same on a 12 KB task, so it should fit, but it is unmeasured); whether libsmb2's `smb2_set_timeout` bounds the TCP connect phase against a silent host (it bounds PDUs — a SYN with no reply may hang longer). A run on a device against a wrong password, a wrong share name and a missing folder would be the useful test.

9 files, +346/−19.

### DIFF
--- a/components/uploader/upload_sched.c
+++ b/components/uploader/upload_sched.c
@@ -752,6 +752,14 @@ void upload_sched_request_reset(void)             { post(EV_RESET, 0); }
 
 void upload_sched_set_busy_fn(upload_sched_busy_fn_t fn) { s_busy_fn = fn; }
 
+bool upload_sched_uploading(void)
+{
+    for (int i = 0; i < s_n_rt; i++) {
+        if (s_rt[i].be && s_rt[i].state == SB_UPLOADING) return true;
+    }
+    return false;
+}
+
 /* ── Progress reporting ───────────────────────────────────────────── */
 
 static const char *state_name(sb_state_t s)
--- a/components/uploader/upload_sched.h
+++ b/components/uploader/upload_sched.h
@@ -83,6 +83,10 @@ void upload_sched_request_reset(void);
 typedef bool (*upload_sched_busy_fn_t)(void);
 void upload_sched_set_busy_fn(upload_sched_busy_fn_t fn);
 
+/* True while any backend is inside a run (connected and transferring), so a
+ * "Test connection" probe does not open a second transport alongside it. */
+bool upload_sched_uploading(void);
+
 /* Compact progress for the Web UI.  Caller frees. */
 esp_err_t upload_sched_progress_json(char **out_json);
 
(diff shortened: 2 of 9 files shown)


ilyakruchinin:

Issue #214, opened 2026-09-06
[FR] - testing SMB/SHQ credentials without saving them first


ilyakruchinin:
#### 1. Upload Scheduler Mutex / Lock
* **Current state:** Uses a check-then-run check via `upload_sched_uploading()`. A scheduled run could theoretically start inside the ~10 s test window.
* **Recommended follow-up:** Add an explicit probe lock or busy flag inside `upload_sched` so that periodic scheduled runs defer while a connection test is in flight, preventing any concurrent TLS/SMB transport contention or double-allocation of buffers.

---

#### 2. Testing Form Inputs without Saving / Rebooting
* **Current state:** Probes settings saved in NVS. In `portal.html`, `saveAll()` triggers a device reboot whenever settings are saved, meaning users must save and wait through a reboot cycle before they can test newly entered credentials.
* **Recommended follow-up:** Update `POST /api/uploads/test-*` to optionally accept credentials in the JSON request body (falling back to NVS if omitted or if set to the password sentinel `••••••••`). This enables on-the-fly credential verification directly from the form without writing to NVS or rebooting.

---

#### 3. Stack Watermark Logging
* **Current state:** `httpd` worker task stack is allocated at 12 KB in PSRAM (`config.stack_size = 12288`). TLS handshakes typically require 6.5–8.5 KB.
* **Recommended follow-up:** Measure actual peak stack consumption on hardware during TLS negotiation by logging `uxTaskGetStackHighWaterMark(NULL)` inside `shq_test()` to confirm stack margin.

---

#### 4. Real Hardware / Negative Case Verification
Validate error reporting on hardware against the following failure scenarios:
* [x] **Wrong SMB credentials:** Verify `smb2_get_error(ctx)` returns a clear authentication failure message.
* [x] **Non-existent SMB share:** Confirm proper share-not-found error message.
* [x] **Missing remote folder path:** Confirm the check identifies that the folder does not exist and prompts the user to create it or update the Remote Path.
* [x] **Unreachable host IP:** Verify network timeout behavior when connecting to a silent/unresponsive host.
* [x] **Invalid SleepHQ Client ID / Secret:** Confirm HTTP 401 error message is returned and displayed in the UI.
* [ ] **No Internet connectivity:** Verify DNS/connection failure message when WAN access is down.

ilyakruchinin:
I'll keep this open for now (let's get some feedback from users before closing, I'll need to do some real hardware tests, too).

[...]

PR #224, opened 2026-09-06
feat(uploader): #214 items 1-3 — test credentials before saving, probe lock, stack watermark

Plantucha:
Items **1**, **2** and **3** of #214. Item 4 (the hardware negative-case matrix) needs a device I do not have.

> **Tested on my fork before opening this.** The same two commits ran through the fork's CI — host tests, static analysis and the mutation check — and all three passed: [green run](https://github.com/Plantucha/SomnoTrace/actions/runs/34049355144). That is not a substitute for your review, and none of it exercises real hardware; it just means the obvious classes of mistake are already ruled out.

---

## 2 · Testing credentials without saving them first — the one users will feel

Testing a connection required saving the form first, and saving reboots the device. So the loop for *"is this password right?"* was: type it, save, wait through a reboot, press Test. The portal said as much out loud — `Save settings first, then test`.

**The root of it was in the backend interface, not the UI.** `test()` read its own settings from NVS, and a backend that loads its own settings can only ever test what is already saved. It now takes the configuration to probe with:

```c
bool (*test)(const uploader_config_t *cfg, char *msg, size_t msg_len);
```

`POST /api/uploads/test-*` accepts an **optional** JSON body of settings, merged over the stored configuration for that one probe. **Nothing is written to NVS** — a connection test must not be able to change what the device is configured to do. No body still means "probe what is saved", so the endpoint's existing contract is untouched.

Three cases the merge has to keep apart, which is most of the work:

| sent | meaning |
|---|---|
| field **absent** | the stored value stands |
| the masked placeholder `••••••••` | also the stored value — an untouched password field shows bullets, and sending them back would probe with a literal row of bullets |
| field **present but empty** | honoured as empty, because clearing a password is how a guest share is tested |

That third row is why "absent" and "empty" cannot be collapsed into one case.

The portal now sends what is on screen and drops the refusal. Passwords are sent only when the field is dirty, so an untouched one is omitted rather than masked-and-sent. The body read is bounded at 1 KB and freed on every path; an unbounded read on the httpd task is how a request becomes a heap exhaustion. Over the bound the body is ignored and the stored settings are used, rather than failing the probe.

## 1 · The probe claims the transport slot instead of checking it

You were right that the window is real. `uploader_test_connection` asked `upload_sched_uploading()` and *then* probed — and the probe takes about 10 s, so a scheduled pass becoming due inside that window would open a second SMB or TLS transport beside it.

`upload_sched_probe_begin()` now tests **and** sets under the same mutex that already guards the runtime slots, and `run_pass()` refuses to start while the claim is held.

Two choices worth your eye:

- **The guard sits in `run_pass()`**, not beside the existing `s_busy_fn` scan check. `run_pass` is the single choke point every route into a run passes through — the periodic tick, `EV_EXPORT`, `EV_SCAN`, manual requests. Guarding only the periodic scan would have left the event-driven paths free to do exactly what the lock prevents.
- **The claim self-expires after 60 s** with a warning. A crash inside a backend's `test()` would otherwise stop every upload silently until the next reboot, and a lock that can wedge the uploader forever is a worse defect than the race it closes.

## 3 · Stack watermark across the TLS handshake

`shq_test` logs `uxTaskGetStackHighWaterMark(NULL)` after the handshake, so the margin is a number rather than an estimate.

⚠️ One trap documented in the code, because it points the wrong way: **ESP-IDF returns this value in bytes; vanilla FreeRTOS returns words.** Scaling by `sizeof(StackType_t)` — the reflex when porting the idiom — would report four times the real headroom, which hides a problem rather than showing one.

---

Built against ESP-IDF v5.5.1: `0x248210`, **43% of the app partition free**. Not flashed — I have no ESP32-S3, so this verifies that it builds and that the logic holds, never that it runs.

`uploader_config_t` gains a struct tag so the vtable can name the type before the definition; both backends take the config by value into a local `cfg`, leaving every existing `cfg.field` reference in them unchanged (~450 bytes on a task with 12 KB).

Item 4's six negative cases — wrong SMB credentials, missing share, missing folder, unreachable host, bad SleepHQ key, no WAN — still want someone with hardware, and the error strings are written to be readable when they fire.

### DIFF
--- a/components/uploader/upload_sched.c
+++ b/components/uploader/upload_sched.c
@@ -103,6 +103,14 @@ static SemaphoreHandle_t s_lock;      /* guards s_rt + s_status for the API */
 
 static upload_sched_busy_fn_t s_busy_fn;
 
+/* "Test connection" probe claim (#214.1), guarded by s_lock like s_rt.  The
+ * timestamp is not decoration: it is what stops a probe that never released
+ * the claim -- a crash inside a backend's test(), a task deleted mid-probe --
+ * from silently stopping every upload until the next reboot.  A lock that can
+ * wedge the uploader forever would be a worse defect than the race it closes. */
+static bool    s_probe_active;
+static int64_t s_probe_since_us;
+
 static int64_t s_next_scan_us;
 static bool    s_scanning;
 static char    s_status[64] = "Starting up";
@@ -530,8 +538,38 @@ static bool run_backend(backend_rt_t *r, int max_days)
 
 /* ── Scheduling pass over all backends ────────────────────────────── */
 
+/* True while a connection test holds the transport claim.  Also the expiry
+ * point: a claim older than UPLOAD_PROBE_MAX_MS is dropped and reported, so the
+ * failure mode of a leaked claim is one warning line and a late upload rather
+ * than an uploader that is permanently and silently idle. */
+static bool probe_in_flight(void)
+{
+    bool held;
+    xSemaphoreTake(s_lock, portMAX_DELAY);
+    if (s_probe_active &&
+        now_us() - s_probe_since_us > (int64_t)UPLOAD_PROBE_MAX_MS * 1000) {
+        ESP_LOGW(TAG, "connection-test claim held for over %d s -- releasing it; "
+                 "an upload pass was deferred until now", UPLOAD_PROBE_MAX_MS / 1000);
+        s_probe_active = false;
+    }
+    held = s_probe_active;
+    xSemaphoreGive(s_lock);
+    return held;
+}
+
 static void run_pass(void)
 {
+    /* ONE choke point, deliberately. Every route into a run -- the periodic
+     * tick, EV_EXPORT, EV_SCAN, a manual request -- arrives through run_pass(),
+     * so guarding here defers all of them.  Guarding only the periodic scan
+     * (as the s_busy_fn hook above does) would leave the event-driven paths
+     * free to open a second transport beside the probe, which is the exact
+     * contention this is meant to prevent. */
+    if (probe_in_flight()) {
+        ESP_LOGD(TAG, "upload pass deferred: a connection test is in flight");
+        return;
+    }
+
     const upload_backend_t *bes[UPLOAD_MAX_BACKENDS];
     int n = uploader_enabled_backends(bes, UPLOAD_MAX_BACKENDS);
     int max_days = uploader_max_days();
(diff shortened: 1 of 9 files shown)


ilyakruchinin:
Thank you @Plantucha! This is a fantastic improvement:

1. **UX win:** Eliminating the "type credentials → save → wait through a reboot → test" loop makes onboarding NAS and SleepHQ significantly smoother.
2. **Concurrency architecture:** Moving from racy check-then-act to atomic test-and-set reservation (`upload_sched_probe_begin()`) under `s_lock` cleanly protects against concurrent TLS + SMB heap exhaustion, and the 60 s auto-expiry prevents any deadlocks if a backend hangs.
3. **Watermark telemetry:** Logging `uxTaskGetStackHighWaterMark(NULL)` on the TLS handshake gives clear, empirical headroom data.

Before we merge, there is one encoding bug on line 1874 of `main/net_provision.c` and one minor HTTP edge-case worth tidying up:

---

### 1. UTF-8 Mojibake in `UPLOAD_TEST_PW_KEEP` (`main/net_provision.c:1874`)

In `main/net_provision.c`:

    /* The eight U+2022 BULLETs the portal shows in a password field it has not been given. Kept as one
     * literal so the sentinel is greppable from both sides of the wire. */
    #define UPLOAD_TEST_PW_KEEP "â€¢â€¢â€¢â€¢â€¢â€¢â€¢â€¢"

* **Root cause:** The literal string was double-encoded as UTF-8 mojibake (`\xc3\xa2\xc2\x80\xc2\xa2` instead of `\xe2\x80\xa2`).
* **Consequence:** If a client or direct API caller sends the literal masked placeholder `••••••••` (as documented in `spec/0011-web-api-endpoints.md`), `strcmp(v->valuestring, UPLOAD_TEST_PW_KEEP)` will not match. The backend will treat `••••••••` as a real password and attempt authentication with eight literal bullet characters instead of falling back to the stored credential.
*(The portal avoids this because `putSecret` omits untouched fields, but any external API call or copy-pasted string will hit this).*
* **Suggested Fix:** Use explicit hex escapes to prevent editor/terminal encoding mismatches across platforms:

    #define UPLOAD_TEST_PW_KEEP "\xe2\x80\xa2\xe2\x80\xa2\xe2\x80\xa2\xe2\x80\xa2\xe2\x80\xa2\xe2\x80\xa2\xe2\x80\xa2\xe2\x80\xa2"

---

### 2. HTTP Socket Drain on Over-Sized Body (`main/net_provision.c:1894`)

In `upload_test_read_overrides`:

    if (len > 1024) return false;

If an incoming request body exceeds 1024 bytes, returning `false` here leaves the unread body on the socket. In ESP-IDF `httpd` with HTTP/1.1 keep-alive enabled, unconsumed body bytes can desynchronize the connection, causing the next request on that connection to read leftover payload as request headers.

* **Suggested Fix:** Either return an explicit error to close the connection:

    if (len > 1024) {
        httpd_resp_send_err(req, HTTPD_413_PAYLOAD_TOO_LARGE, "request body too large");
        return false;
    }

or drain the socket before returning if you prefer falling back to stored settings.

---

Once line 1874 is adjusted, I'm very happy to merge this!

Plantucha:
Fixed in `c9237a8` — thank you, that was a real bug and mine.

**1. The sentinel.** I checked the bytes before changing anything: the literal was 48 bytes, eight copies of `c3 a2 c2 80 c2 a2`, with zero occurrences of `e2 80 a2`. Exactly as you diagnosed. It is now spelled as hex escapes, and I verified it in the *built image* rather than only in the source — the binary carries `e2 80 a2` ×8 and none of the old sequence.

**2. The over-sized body.** Implemented, with one change to where the check lives. Calling `httpd_resp_send_err()` inside `upload_test_read_overrides()` would send two responses: the helper returning `false` means "no overrides — fall back to stored config", and the caller then carries on to `uploader_test_connection()` and sends its normal JSON body. So the bound moved up into `upload_test_send()`, which answers 413 and returns `ESP_FAIL` — that closes the socket, so the undrained body cannot desynchronise the next request on a keep-alive connection. The `1024` is now `UPLOAD_TEST_BODY_MAX` instead of living in two places.

One small thing worth knowing: `HTTPD_413_PAYLOAD_TOO_LARGE` is not in ESP-IDF v5.5.1 — the enum is `HTTPD_413_CONTENT_TOO_LARGE` (`esp_http_server.h:615`). That is what I used.

Builds clean in the pinned v5.5.1 container.

ilyakruchinin:
Merged!

JONESN22:
Sent from my iPadOn 9 Sep 2026, at 00:55, Ilya Kruchinin ***@***.***> wrote:﻿ilyakruchinin left a comment (ilyakruchinin/SomnoTrace#224)
Merged!

PR #229, opened 2026-09-08
ci: static analysis for the firmware sources and this repo's shell scripts

Plantucha:
Static analysis for the firmware sources and this repo's shell scripts, in the tiered shape agreed in discussion #219.

## What it does

**Blocking tier** — `cppcheck` `error,warning,performance,portability` over `main/` and `components/`, plus `shellcheck --severity=warning` on the repo's own scripts. Green on `main` today, so it starts clean rather than arriving with a backlog to burn down.

**Advisory tier** — `cppcheck --enable=style` runs and prints a per-check summary but never fails the job. Measured on current `main`: **126 style findings**, led by `constVariablePointer` (55), `constVariable` (17) and `constParameterCallback` (14). Useful to browse, not worth blocking a merge on.

**Excluded** — `third_party/`, `build/`, `managed_components/`.

## The two suppressions, and why they are inline

Both are false positives, both are documented at the site:

- `main/net_provision.c` — the linker-symbol subtraction you already knew about. cppcheck calls it `comparePointers` at 2.13 and `subtractPointers` at 2.19, so the directive names **both** IDs.
- `components/uploader/upload_sched.c` — a `uninitvar` reported at the *read* site rather than the call.

They are inline rather than in a suppressions file because cppcheck 2.13 — the version `ubuntu-latest` ships — rejects comments in a suppressions list (`Failed to add suppression. No id`), and because a suppression that lives next to the code it excuses is the one that gets deleted when the code changes.

## Version drift

`scripts/lint-versions.env` records the versions the tiers were measured against. The workflow prints what the runner actually has and emits a `::warning::` when they differ — a descriptive pin, not a hard one, so an image bump surfaces as a note rather than a red build. This matters more than it sounds: 2.13 and 2.19 disagree on both check IDs and finding counts, which is how the first version of this landed red.

## `scripts/lint.sh`

The same gate a human can run locally, so nobody has to read the YAML to reproduce a CI failure. `--docker` for a pinned container if the host tools differ.

## Testing

Ran on the fork against `main` at `bca2236`: **blocking tier clean**, advisory 126. The CI job itself has run green on the fork.

Measurements and the rejected candidates (`clang-tidy`, `codespell`) are in #219 and #220.


ilyakruchinin:
Merged!

PR #230, opened 2026-09-08
ci: build the release images in CI and attest them

Plantucha:
Builds the two distributable images in CI and attests them, as proposed in discussion #223.

## What it does

On a `v*` tag (or `workflow_dispatch`) it runs the repo's own `scripts/build-dist.sh`, checks that both images exist and are plausibly sized, attests them with `actions/attest-build-provenance@v2`, and uploads them as workflow artifacts.

## The three points you raised in #223

**1. Verification.** After a release built this way, anyone can run:

```
gh attestation verify somnotrace-<ver>-ota.bin --repo ilyakruchinin/SomnoTrace
```

Measured on the fork: exit 0 against the correct repo, 404 against the wrong one, and 404 again after flipping a single bit in the image. That last one is the property that matters — the attestation is over the artifact digest, so a modified binary cannot inherit it.

**2. The release boundary stays manual.** There is no publish step in this workflow. The only output is `actions/upload-artifact`, so a tag produces attested images for you to look at and attach yourself. Nothing reaches the releases page without you putting it there.

**3. The repository guard, as you asked:**

```yaml
jobs:
  release:
    if: github.repository == 'ilyakruchinin/SomnoTrace'
```

Without it a fork pushing a `v*` tag would build, attest and upload a pair of official-looking images. The attestation would still be honest — it names whichever repository produced it, which is the whole point — but the artifacts would exist, and a maintainer triaging a report has no reason to expect that.

## One fix folded in

`ci(release): resolve the real images, not the latest-* symlinks` — `build-dist.sh` leaves `latest-*` symlinks beside the versioned images, and globbing picked them up, so the attestation subject could end up being a symlink rather than the file. Found on the fork after the original proposal.

## Cost

Seconds, not minutes, and free on a public repository. `permissions:` is scoped to `contents: read` plus the `id-token: write` / `attestations: write` the action requires.

## Testing

Run on the fork end to end: images built, attested, verified with `gh attestation verify`, and the negative controls above. The guard means this job will not run on the fork again, which is the intended behaviour.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,13 +39,20 @@ jobs:
           ./scripts/build-dist.sh 2>&1 | tee release-build.log
           ls -l dist/
 
-      - name: Confirm both images exist and are plausible
+      - name: Confirm both images exist, and name them for attestation
         # A release step that exits 0 having published nothing is the failure this
         # catches. Sizes are a floor, not a checksum: an empty or truncated image is
         # the realistic accident, not a subtly wrong one.
+        #
+        # -type f is the whole trick. build-dist.sh also writes
+        # somnotrace-latest-full.bin as a SYMLINK to the versioned file, so a plain
+        # dist/*-full.bin glob matches two paths for one artifact — which failed this
+        # check on its first run, correctly. Attesting the symlink as well would have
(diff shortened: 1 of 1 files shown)


ilyakruchinin:
Merged!

PR #231, opened 2026-09-08
feat: stamp the source repository into the build and show it in the portal

Plantucha:
The technical half of discussion #222 — the build stamp and the Status row. **`NOTICE` is untouched**, per your decision there.

## The problem

A screenshot of the Status tab cannot currently tell you which repository the firmware was built from. A bug report from a fork's build looks exactly like one from an official image, and gets triaged against source that does not contain the bug.

## Two values, deliberately different in kind

**`SNT_SOURCE_REPO`** — read from `GITHUB_REPOSITORY` at configure time. A fork building through its own CI stamps its own slug without anyone remembering to, because GitHub Actions sets that variable for every run in every fork and it cannot be inherited by accident. Off CI it reads `local build`.
*Answers: where was this built?*

**`SNT_UPSTREAM_REPO`** — a constant in the tree, not from the environment, so it travels with the code and a fork would have to delete it on purpose. This one exists because the source slug alone leaves a reader looking at a repository name they have no reference point for.
*Answers: what is this a fork of?*

## What the user sees

```
Firmware     2.0.3
Built from   ilyakruchinin/SomnoTrace
Uptime       ...
```

The **Upstream** row renders only when the build came from CI *and* the two slugs differ. On an official build they are identical and a second line is just noise; a local build is not a fork and must not be labelled as one. So official images look exactly as they do today.

## Not an authenticity control

The source comments say this outright, in both `CMakeLists.txt` and `net_provision.c`. Any string compiled into a binary can be copied verbatim by whoever rebuilds it. This makes honest builds legible, which is a triage aid — it is not evidence of origin. Proving origin needs a signature over the artifact, which is the separate proposal in #223.

## Cost

+1.3 KB. Three files, +53 lines, no behaviour change to anything that exists.

## Testing

Builds clean in the pinned IDF v5.5.1 container. Verified in the built image rather than only in source: `local build` present off CI, `ilyakruchinin/SomnoTrace` present as the upstream constant, and the Upstream row correctly hidden when the two match.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,6 +85,34 @@ endif()
 
 set(PROJECT_VER "${FIRMWARE_VERSION}")
 
+# ── Source repository stamp ─────────────────────────────────────────────────
+# WHICH REPOSITORY THIS BINARY WAS BUILT FROM, decided at build time rather than
+# written into the source. A fork that builds through its own CI stamps its own
+# slug without anyone having to remember to: GITHUB_REPOSITORY is set by GitHub
+# Actions for every run, in every fork, and cannot be inherited by accident.
+#
+# ⚠️ THIS IS NOT AN AUTHENTICITY CONTROL and must never be presented as one. Any
+# string compiled into a binary can be copied verbatim by anyone rebuilding it.
+# It answers "where did this build come from?" honestly for honest builds, which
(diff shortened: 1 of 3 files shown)


ilyakruchinin:
Merged!

PR #232, opened 2026-09-08
test(tools): a mutation harness that searches, alongside the one that pins

Plantucha:
`scripts/mutants.py`, now running in CI as of #221, carries 18 hand-written mutants at sites someone chose. That is a **regression guard**: it proves those specific behaviours stay asserted. By construction it cannot find a gap nobody thought of.

This is the other half — and it is meant to sit next to that one, not replace it.

## What it does differently

It applies relational and arithmetic operators systematically across a whole file, then classifies every survivor using gcov line coverage:

| | meaning | fix |
|---|---|---|
| **UNREACHED** | no test executes the line | needs a test |
| **UNASSERTED** | a test runs it and passes anyway | needs a stronger assertion |

That split is the reason to have it. "Survivor" on its own does not tell you which of the two problems you have, and they have different fixes.

**Headers are targets**, not only the `.c` files a test links — `spool_to_edf()` lives in `main/edf_data_dict.h`, and that is the code #202 was about.

## What stops it lying

**A canary per file.** Before mutating, it empties a function body and requires the suite to catch it. A harness that cannot kill an emptied function is measuring nothing and would otherwise report a clean sheet.

**`--selftest`** plants one killable and one equivalent mutant and asserts it catches exactly the first:

```
baseline                     PASS
planted KILLABLE mutant      caught (correct)
planted EQUIVALENT mutant    survived (correct)
VERDICT: harness discriminates
```

**It refuses rather than guesses.** If the baseline does not pass on the unmutated tree it prints `VOID — no verdict below would mean anything` and stops. That fired for real while preparing this PR: run against an older checkout, two suites were skipping for want of cJSON and the "results" were meaningless. The harness said so instead of reporting them.

**`scripts/mutate_equivalent.txt`** records mutants no input can distinguish from the original, with the argument for each. An entry there *suppresses* a result, so the file states plainly that "I could not think of an input" is not the same claim as equivalence.

## Results on `main` today

```
main/as11_time.c        killed 20  unasserted 0  unreached 0  known-equivalent 4
main/edf_data_dict.h    killed  2  unasserted 0  unreached 0  known-equivalent 3
main/oximetry_vld3.c    killed 20  unasserted 0  unreached 1  stillborn 4
                        0 unasserted survivors
```

The one unreached line is `oximetry_vld3.c:58` — a `<` → `<=` on the header-length bound that no test reaches.

**Cost: 35 s** for the full sweep. For comparison `mutants.py` takes 61 s in CI and #218's IDF build 3 m 21 s.

## One thing I changed, and one question for you

The harness originally found cJSON via `$SNT_CJSON` or a bounded glob. It now also honours **`$CJSON_DIR`**, which `scripts/run_host_tests.sh` already documents — so a checkout that can run the host suite can run this without learning a second variable.

**This PR does not wire it into CI.** With one unreached line it would need either a new test or an accepted baseline on day one, and that is your call rather than mine. Merged as a tool it costs nothing and is there when someone wants to ask whether a test actually asserts what it appears to. If you would rather it gated, I will add the workflow and the test for `oximetry_vld3.c:58` in a follow-up.

### DIFF
--- a/scripts/mutate_equivalent.txt
+++ b/scripts/mutate_equivalent.txt
@@ -1,0 +1,35 @@
+# Mutants that NO input can distinguish from the original — unkillable by anyone, and therefore not
+# gaps. Recorded with the reasoning so the next run does not re-report them and the argument outlives
+# whoever worked it out.
+#
+# Format:  <path>:<line>:<operator>      # why it is equivalent
+#
+# ⚠️ An entry here SUPPRESSES a result. Add one only after showing the two versions cannot differ —
+# "I could not think of an input" is not the same claim. When in doubt, leave it out: a re-reported
+# survivor costs a minute, a wrongly suppressed one hides a defect for as long as the entry stands.
+
+# Hinnant civil-date arithmetic: the sign branches differ from their mutants only at year 0
+# (y == 0 after the March-shift; z == 0 days since 0000-03-01). No AS11 timestamp reaches there.
+main/as11_time.c:58:relational:>=→>
+main/as11_time.c:68:relational:>=→>
+# spool_to_edf: each mutant differs from the original only at a value where both paths agree.
+main/edf_data_dict.h:49:relational:>=→>     # prod == 0: (0+half)/den == (0-half)/den == 0 since half < den
+main/edf_data_dict.h:50:relational:>→>=     # rounded == INT16_MAX: early return and cast both yield INT16_MAX
+main/edf_data_dict.h:51:relational:<→<=     # rounded == INT16_MIN: same, at the other end
+
+# as11_time.c — survivors of the 2026-09-04 --all run that no input can separate from the original.
+main/as11_time.c:85:relational:<→<=        # floor_div: (a < 0) vs (a <= 0) differ only at a == 0, and a == 0
+#   makes a % b == 0, so the && short-circuits before either form is evaluated.
+main/as11_time.c:111:equality:!=→==        # set_offset: the branch holds only an ESP_LOGI and (void)mins. On host
+#   the log is a no-op; on target the two forms differ only in WHEN the "AS11 UTC offset" line is
+#   logged (changed vs unchanged), never in the stored offset. Not testable, not a defect.
+main/as11_time.c:184:arithmetic:+→-        # period_start: k runs over {-1, +1}, so raw + k*86400 and raw - k*86400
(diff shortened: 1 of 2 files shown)


Plantucha:
Pushed `9ac1971` — a correctness fix to the harness, and it changes a number in the description above. Flagging it rather than letting you find it.

**The bug.** The reach parser treated three gcov markers as "not executed": `#####`, `=====` and `-`. The first two are right. `-` is not — it means the line carries no executable code *of its own*, which is what gcov prints for the continuation lines of a multi-line expression:

```
      20:   57:    if (!data || !out || len < OX_VLD3_HEADER_LEN ||
       -:   58:        source_size < OX_VLD3_HEADER_LEN + OX_VLD3_RECORD_LEN)
```

The whole condition is charged to line 57; line 58 gets `-`. So a mutant on 58 was reported UNREACHED — *"write a test that reaches it"* — for a line executed twenty times. Worse, a test that genuinely killed that mutant could not clear the report, which is how I found it: I wrote the test, verified by hand that it kills the mutant, and the harness still said UNREACHED.

**Why it matters more than a miscount.** UNREACHED and UNASSERTED have different fixes, and the split between them is the only reason to prefer this over a plain kill count. Sending someone to write a test they do not need, for a line already covered, is worse than saying nothing. Every wrapped condition in the tree was affected — which is most of the interesting ones.

**The correction to the numbers above.** Same tree, before and after:

```
before   killed 20  unasserted 0  unreached 1     <- the 1 was false
after    killed 20  unasserted 1  unreached 0     <- and the 1 is real
```

So `main` does have one genuine gap in `oximetry_vld3.c`, and it is an unasserted survivor rather than an unreached line — the opposite diagnosis. I have opened that fix separately as a one-file test change; with it applied the harness reads `killed 21, unasserted 0, unreached 0`.

`--selftest` still discriminates. Full sweep 42 s.

**On the CI question in the description:** this removes my own reservation. The reason I said "not wired into CI by this PR" was that one unreached line would need a baseline on day one. That line was an artefact. With the test PR merged alongside, the sweep is clean and could gate honestly — still your call, and I am happy either way.

Same shape as the `N*` bug the comment in that function already described: a gcov form the parser did not understand, read as absence. The rule is now the narrow one — only `#####` and `=====` mean never-executed; everything else is kept and run.

ilyakruchinin:
Merged!

Issue #88, opened 2026-08-06
Cross-noon Theraphy


irfanguder:
Today I have run a therapy between 10:15 and 14:00 to test how the system behaves if a therapy runs through noon.

AutoSync SHQ;
5 August
![image](https://github.com/user-attachments/assets/1d57ee13-de4c-45db-8489-41f3be567f6c)

6 August
![image](https://github.com/user-attachments/assets/013ade34-e73c-4629-bf31-b176d296edfb)

SomnoTrace SHQ;
5 August 
![image](https://github.com/user-attachments/assets/9eaf52e5-b79c-4dd6-af67-85001aa416ae)

I don't know which approach is the correct one, also it is not a serious problem but I just wanted to share my finding for your considerations.
Regards.

vvkovmDdjhrd:
Dear @irfanguder, your graphs seem to be at different scales (and drift, of course). To understand how to proceed, I would make them all scale the same. 

Is it also expected that ST should close the data transmission session at 12:00 and open a new one for the new date?

Best Regards.

ilyakruchinin:
The cross-noon border is an interesting case.
I know AS11 does special treatment of this use case (splits therapy), and I haven't looked into it.

Other than the non-split graph - any issues with SomnoTrace exported files @irfanguder ?

ilyakruchinin:
I'm just wondering if it presents an actual issue.
I understand it is different from what AS11 does, but need to get more details to understand if it's worth fixing or is that actually a "feature"/improvement over AS11 behaviour.

irfanguder:
I am not sure what the proper approach for a cross-noon therapy should be but I have tested and shared the results.

For me it is not an issue or even a feature request but if there is a standard for this use case we can implement it.

Other than date split I didn't see any problem in SomnoTrace.

ilyakruchinin:
Thank you!
I'll give it some thought.
I know we should probably stay as close to AS11- generated files as possible.

beren12:
I would just change the timezone on the machine so the middle of your sleep is midnight, if you sleep through noon…

ilyakruchinin:
Looks like this one might actually be a feature, not a bug (yes, that's a deviation from AS11-behaviour, but it might actually be desirable for those sleeping through the noon-mark).

beren12:
[...]

PR #234, opened 2026-09-08
docs: describe cross-noon sessions as the feature they are

Plantucha:
The documentation half of #88. The clamp that makes cross-noon sessions safe landed in `cce8337`; this is the part that tells a user about it.

Your scoping on that issue was *"Maybe some edge-case analysis to ensure it gets displayed properly (in all platforms — internal UI, OSCAR, SHQ) + documenting it as a feature."* The analysis happened in the thread; the documenting did not, and the issue is still open because of it.

## What it adds

A `<details>` block in **Advanced Features & Documentation**, alongside the FTP one, plus a row in the Feature Overview table.

It covers:

- that therapy spanning midday is recorded as **one continuous session**, and that the AS10/AS11 cannot do this
- why the noon boundary exists at all — ResMed's noon-to-noon day, so a session starting before 12:00 belongs to the previous date
- what the session looks like in the dashboard, in OSCAR, and in the exported EDFs
- why OSCAR shows it as one session: it takes session extent from the recording files, not from the `MaskOn`/`MaskOff` pairs, which only group files and validate the day
- why a window past 1440 is **clamped** rather than dropped — a dropped STR record leaves the night importable but strips its settings (mode, pressures, EPR) and the day's statistics

## What it deliberately does not claim

Two limits are stated in the text rather than left for a reader to assume:

- OSCAR's behaviour is read from `resmed_loader.cpp`, not measured against a running copy
- SleepHQ's importer is closed, so nothing is claimed for it

I would rather the README be explicit about which of the three platforms was actually verified than imply all of them were.

Documentation only — no code touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,33 @@ Connect with any FTP client (e.g., FileZilla) to `ftp://somnotrace.local`.
 
 </details>
 
+<details>
+<summary><b>🕛 Sessions That Cross Noon</b></summary>
+
+Therapy that runs through midday — a late morning lie-in, a nap that starts before 12:00 and ends after — is recorded by SomnoTrace as **one continuous session**. The AS10 and AS11 cannot do this: they cut a session at noon and write the two halves to different days.
+
+This is deliberate, and it is why SomnoTrace records the stream itself rather than relying on the machine's own summary.
+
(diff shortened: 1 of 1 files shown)


ilyakruchinin:
Merged!

Issue #151, opened 2026-08-27
SHQ Upload: sleephq: finalise failed


irfanguder:
For the last couple of days I am getting this error during SHQ uploads, since I am on holiday with my AS10 the tests are executed with my previous sessions. I have recreated the EDFs but it didn't help.
When I checked the import list in SHQ some of the sessions are displayed as imported.
Repetitively 
GET /api/v1/imports/16012359 
upload_shq: response: HTTP 200 lines appear in the logs.
This is not related with the latest release I have the same error in the previous release as well, the latest successful upload to SHQ was on Saturday. After that since yesterday I haven't tried SHQ     Upload. 

Not an urgent case but for your evaluation,

![image](https://github.com/user-attachments/assets/018e73b1-ab80-4f58-a8e6-b6239b0d5281)

![image](https://github.com/user-attachments/assets/c441ce0b-c64e-4b6a-8688-75f8be86344c)

17:29:36.014) upload_shq: response: HTTP 201 (889 bytes body)
I (17:29:36.015) upload_shq:   uploaded CurrentSettings.json (2182 bytes, hash=7bc8d8b217b86ef8e676264c7258a404)
I (17:29:36.255) upload_shq: response: HTTP 201 (891 bytes body)
I (17:29:36.256) upload_shq:   uploaded CurrentSettings.crc (4 bytes, hash=5b4de589651d1d7f3d1a817ff4131b76)
I (17:29:36.256) upload_shq:   root bundle: 5 file(s)
I (17:29:36.257) upload_shq: processing import 16012359...
I (17:29:36.257) upload_shq: sending POST /api/v1/imports/16012359/process_files (200 bytes)
I (17:29:36.449) upload_shq: response: HTTP 201 (13 bytes body)
I (17:29:36.450) upload_shq: import 16012359 processed
I (17:29:36.450) upload_shq: sending GET /api/v1/imports/16012359 (185 bytes)
I (17:29:36.673) upload_shq: response: HTTP 200 (1456 bytes body)
I (17:29:38.668) upload_shq: sending GET /api/v1/imports/16012359 (185 bytes)
I (17:29:38.872) upload_shq: response: HTTP 200 (1456 bytes body)
I (17:29:40.868) upload_shq: sending GET /api/v1/imports/16012359 (185 bytes)
I (17:29:41.058) upload_shq: response: HTTP 200 (1455 bytes body)
[...]

PR #235, opened 2026-09-08
fix(sleephq): say WHY a finalise failed, instead of one message for three causes

Plantucha:
`shq_wait_import()` returns a bare `ESP_FAIL` from three different places:

- a non-2xx reply to the status query
- a body `cJSON_Parse` rejected
- an import SleepHQ actually marked `failed`

The caller prints the same sentence for all three — `import processing failed for day <d>` — which reads as *"SleepHQ rejected the import"* whichever one happened. #151 has been open since 27 August with logs that cannot separate our bug from their verdict.

## What this adds

Each path now says which it was. The rejected case prints SleepHQ's status string verbatim. The unparseable case prints the body length and its first 120 bytes, which separates the two ways that can happen: a truncated body shows a length at the read cap, a de-chunking fault shows hex chunk prefixes in the text. The poll loop says so when it exhausts its thirty attempts.

## No behaviour change

Deliberately. I have no SomnoTrace hardware, so this had to be safe to merge without me having run it on a device.

The same conditions return the same values in the same order. The only structural change is that `free(body)` moves after the parse check so the body can be logged before it is released. The absent-status warning fires on the first poll only, because that case still retries and would otherwise log thirty times.

I did write a behaviour change into an earlier draft — making an absent status fail immediately instead of retrying — and took it out. It may well be right, but it is a separate decision and not one to smuggle into a logging commit.

## What the log already tells us, and what it cannot

Measured against [the log attached to #151](https://github.com/user-attachments/files/31722757/somnotrace-logs.1.txt) (v1.4.1, day 20260807):

```
08:10:27.689  response: HTTP 201 (13 bytes body)          <- process_files accepted
08:10:27.690  sending GET /api/v1/imports/16117550
08:10:28.183  response: HTTP 200 (1971 bytes body)
08:10:28.187  import processing failed for day 20260807
```

Four milliseconds, on the **first** poll of thirty.

Worth recording a negative here: I expected our read buffer to be truncating. It is not. `shq_http_read_response()` allocates `SHQ_RESP_CAP + 1024` = 5120 bytes and stops with a bare `break` when full — a real latent hazard, but not this. The failing body is 1971 bytes and the largest anywhere in that log is 1977. The body arrives whole, and the failure is decided in the lines this PR instruments.

Builds clean in the pinned IDF v5.5.1 container.

Refs #151.

### DIFF
--- a/components/uploader/uploader_sleephq.c
+++ b/components/uploader/uploader_sleephq.c
@@ -562,20 +562,52 @@ static esp_err_t shq_wait_import(esp_tls_t *tls, const char *import_id)
         char *body = NULL; size_t body_len = 0;
         int status = shq_http_request(tls, "GET", path, NULL, s_token,
                                        NULL, NULL, &body, &body_len);
-        if (status < 200 || status >= 300) { free(body); return ESP_FAIL; }
-        cJSON *root = cJSON_Parse(body); free(body);
-        if (!root) return ESP_FAIL;
+        /* Every failure below used to return the same bare ESP_FAIL, and the caller
+         * reports all of them as "import processing failed for day <d>" -- which reads
+         * as "SleepHQ rejected the import" whichever one actually happened. Three
+         * different causes wearing one message is why #151 could run for weeks without
+         * the cause being identifiable from a log. Say which.
+         *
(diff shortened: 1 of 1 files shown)


ilyakruchinin:
Merged!

PR #249, opened 2026-09-09
ci(lint): analyse the build, not the directory — and every board, not one

Plantucha:
Four follow-ups to #229, all found by pointing the gate at a large real change rather than at a clean tree. The headline: **as merged, the gate finds 1 finding in a 92,000-line branch. After these, it finds 15.**

## 1 · The pointer-subtraction suppression does not scale

`EMBED_FILES` gives every blob a `_binary_<name>_start[]` / `_binary_<name>_end[]` pair, and `end - start` is how you get its length. cppcheck calls it undefined behaviour, every time. Suppressing that at the one site we had was fine while there was one site — but a **new** false positive appears for every blob anyone embeds, so the gate goes red on correct code and whoever hits it has to know this history to get past it.

Measured: on that branch the blocking tier reported four errors, two of them this pattern in `main/timezone_catalog.c`, a file that did not exist when the suppression was written.

Now suppressed tree-wide, with what it gives up written in the script rather than buried: a genuine subtraction across two different objects would no longer be caught. Every instance in this tree is the `_binary_*` shape, so the check has found nothing else.

## 2 · shellcheck had one tier and it was blocking

cppcheck was split blocking/advisory; shellcheck was not, and the asymmetry was not deliberate. On fifteen real scripts it produced 8 blocking findings — **7 of them SC2034** on one shape:

```bash
IFS=$'\t' read -r BOARD BUILD_DIR SDKCONFIG DEFAULTS BUILD_REQUEST CLEAN_REQUEST <<< "$P"
```

You cannot read positional fields without naming the ones you do not use, so any record-parsing script trips it forever, for correct code. SC2034 moves to advisory; everything else at `--severity=warning` stays blocking. The eighth finding was SC2164 — a `cd` with no `|| exit` — which is exactly what the gate is for and is fixed rather than suppressed.

The new advisory shell tier also surfaces 8 findings on this branch that nothing was printing before.

## 3 · The gate analysed text, not the build

This is the one that matters. `cppcheck main components` reads source as text — no `-D`, no include paths, no way to know which branch of an `#if` is live. On a project whose file list is chosen by Kconfig that is not a weaker check, it is a **different** one: it reads code the compiler never sees and misses code the compiler does.

```
directory scan       1 finding
compile database    15, of which 14 new to that branch
```

Among the 14: an out-of-bounds read of a 7-element array guarded by `< 8`, an uninitialised variable, a null dereference on allocation failure. A text scan cannot reach any of them, because those files are not in the default configuration at all.

`idf.py reconfigure` emits the database without compiling, so this costs seconds. Three details each cost a debugging round and are commented in place: the database carries the container's `/project/...` paths; unfiltered it also carries ~1,100 ESP-IDF translation units; and `--project` **must** be absolute or the path-glob suppressions silently stop matching and the gate goes red on LVGL.

If no database can be produced — no Docker, no prior build — it falls back to the directory scan and says so in three lines, rather than reporting a clean tier it did not earn.

## 4 · One database was still one board

The commit above claimed 15 findings. On a clean checkout it found **3**. I had measured it in a tree where an earlier manual build left an sdkconfig with a second board enabled, and `reconfigure` picked it up — I wrote "one database is one configuration" into that commit and then missed that my own measurement had broken the rule.

The gate now enumerates `sdkconfig*.defaults` and lints each configuration, unioning findings. `SDKCONFIG_DEFAULTS` is only read when no sdkconfig exists yet, so each leg gets its own `SDKCONFIG` path — without that, four legs silently resolve to the same board while reporting four names.

## Cost

This repo has one `sdkconfig*.defaults`, so it is one configuration and one reconfigure: **12 s, blocking tier clean, 124 style findings advisory**. It stops being a no-op the moment a second board lands, which is when a single-configuration gate starts reporting a pass it did not earn.

### DIFF
--- a/.github/workflows/host-tests.yml
+++ b/.github/workflows/host-tests.yml
@@ -33,6 +33,26 @@ jobs:
 
       - name: Install libcjson
         run: |
+          # ⚠️ DROP THE RUNNER'S UNRELATED APT SOURCES FIRST. The ubuntu-latest image ships
+          # apt sources for Google Chrome and Microsoft that this project never installs from.
+          # When one of them serves a bad index, `apt-get update` exits 100 and the job dies
+          # before reaching a line of project code. Measured 2026-09-09: `Hash Sum mismatch`
+          # on dl.google.com/linux/chrome-stable reddened host-tests, lint and mutation
+          # together, and survived a re-run — an outage, not a flake.
+          #
+          # MATCHED BY CONTENT, NOT BY FILENAME. The first version of this named
+          # /etc/apt/sources.list.d/google-chrome.list, removed nothing, and failed exactly as
+          # before: the image's file names differ between runner releases and between the
+          # .list and deb822 .sources formats. A guessed name fails silently, which is the
+          # worst way for a workaround to be wrong. Grepping for the host cannot miss.
+          #
+          # Removed rather than swallowed with `|| true`, which would also hide a genuine
+          # failure to reach the Ubuntu archive — the one repository these jobs depend on.
+          for f in $(grep -rlE 'dl\.google\.com|packages\.microsoft\.com' \
+                       /etc/apt/sources.list /etc/apt/sources.list.d/ 2>/dev/null || true); do
+              echo "removing apt source this project does not use: $f"
+              sudo rm -f "$f"
+          done
           sudo apt-get update -q
           sudo apt-get install -y --no-install-recommends libcjson-dev
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -53,6 +53,19 @@ jobs:
         # the drift is reported below, which is what makes an unexplained new finding
         # distinguishable from a regression the contributor actually introduced.
         run: |
+          # ⚠️ DROP THE RUNNER'S THIRD-PARTY APT SOURCES FIRST. The ubuntu-latest image ships
+          # lists for Google Chrome, Microsoft and Azure that this project never installs from.
+          # When one of them serves a bad index, `apt-get update` exits 100 and every job that
+          # installs anything goes red before reaching a line of project code. Measured
+          # 2026-09-09: `Hash Sum mismatch` on dl.google.com/linux/chrome-stable reddened
+          # host-tests, lint and mutation together, and survived a re-run — it is an outage,
+          # not a flake, and retrying does not clear it.
+          #
+          # Removing the lists is the fix rather than `|| true`, which would hide a genuine
+          # failure to reach the Ubuntu archive — the one we do depend on.
+          sudo rm -f /etc/apt/sources.list.d/google-chrome.list \
+                     /etc/apt/sources.list.d/microsoft-prod.list \
+                     /etc/apt/sources.list.d/azure-cli.list
           sudo apt-get update -q
           sudo apt-get install -y --no-install-recommends cppcheck shellcheck
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -53,19 +53,26 @@ jobs:
         # the drift is reported below, which is what makes an unexplained new finding
         # distinguishable from a regression the contributor actually introduced.
         run: |
-          # ⚠️ DROP THE RUNNER'S THIRD-PARTY APT SOURCES FIRST. The ubuntu-latest image ships
-          # lists for Google Chrome, Microsoft and Azure that this project never installs from.
-          # When one of them serves a bad index, `apt-get update` exits 100 and every job that
-          # installs anything goes red before reaching a line of project code. Measured
-          # 2026-09-09: `Hash Sum mismatch` on dl.google.com/linux/chrome-stable reddened
-          # host-tests, lint and mutation together, and survived a re-run — it is an outage,
-          # not a flake, and retrying does not clear it.
+          # ⚠️ DROP THE RUNNER'S UNRELATED APT SOURCES FIRST. The ubuntu-latest image ships
+          # apt sources for Google Chrome and Microsoft that this project never installs from.
+          # When one of them serves a bad index, `apt-get update` exits 100 and the job dies
+          # before reaching a line of project code. Measured 2026-09-09: `Hash Sum mismatch`
+          # on dl.google.com/linux/chrome-stable reddened host-tests, lint and mutation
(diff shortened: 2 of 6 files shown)


ilyakruchinin:
Linting failed

Plantucha:
Fixed in `2fcd117` — thank you, and sorry for the noise. Two defects, both mine.

**The suppression.** I removed the inline `cppcheck-suppress` directives from `net_provision.c` in favour of a tree-wide `--suppress=comparePointers`, which passed on my machine and failed here with exactly the finding it was meant to silence. cppcheck **2.13 does not apply an id-only `--suppress` to findings produced from a `--project` run**; 2.19 does. I had validated a version-sensitive change against 2.19 — and `scripts/lint.sh` ships a `--docker` mode for precisely this, which I did not use. Both forms are now present, with a comment saying why removing either one reds the gate.

**And a second one the first was hiding.** Running `--docker` to reproduce, the tier failed with `0 script(s)`: inside the container the checkout is bind-mounted, `git ls-files` returns nothing, and shellcheck exited non-zero for having no input. So the gate went red while naming no finding — worse than not gating. There is now a `find` fallback and a warning if the list is still empty.

Verified both ways this time: `--docker` (2.13, falls back to the directory scan and says so) exit 0, and native (2.19, compile database, 40 translation units) exit 0. Blocking tier clean in both.

ilyakruchinin:
Thanks for this, @Plantucha! Analyzing the compile database rather than doing a text scan is structurally the right model for an ESP-IDF codebase with Kconfig-guarded sources. Catching the out-of-bounds array read on the OTA failure path is concrete proof of its value.
Before merging, there are a few edge cases and architectural points that should be addressed:
#### 1. Disconnect between `scripts/lint.sh` and CI (`.github/workflows/lint.yml`)
* **Issue:** This PR updates `scripts/lint.sh`, but `.github/workflows/lint.yml` was not updated. The CI workflow currently calls `cppcheck main components` directly (the text scan). As a result, CI is still running the old check and does not benefit from the compilation database; only developers running `./scripts/lint.sh` locally will run it.
* **Recommendation:** Update `.github/workflows/lint.yml` to use `scripts/lint.sh` (or align its steps) so that PRs in CI actually run the compile-database gate. Note that generating `compile_commands.json` requires Docker on the runner (`./scripts/lint.sh` invokes `scripts/idf.sh ... reconfigure`), which is available on `ubuntu-latest`.
#### 2. Naive `/project` substring replacement in Python
* **Issue:** In `compile_db_for()`:
      f = e["file"].replace("/project", root)
      cmd = cmd.replace("/project", root)
  Python's `str.replace()` replaces *all* occurrences of the substring. If a developer checks out the repository in a directory containing `/project` (e.g. `/home/user/projects/SomnoTrace`), or if any component/source file has "project" in its name (e.g. `main/net_project.c`), paths and compiler flags will be mangled.
* **Recommendation:** Use anchored prefix replacement for file paths and directory targets:
      if f.startswith("/project/"):
          f = root + f[len("/project"):]
  And for compiler flags in `cmd`, ensure replacements match whole path boundaries or whitespace/quotes.
#### 3. Docker file ownership on `$bdir` (`build-lint-$n_cfg`)
* **Issue:** When `./scripts/idf.sh ... reconfigure` runs, Docker runs as root (or the container UID), creating `build-lint-$n_cfg/` owned by root. On Linux host systems without POSIX directory ACL inheritance, the subsequent `rm -rf "$bdir"` executed by the host user can fail with `Permission denied`.
* **Recommendation:** Consider handling cleanup inside the container if needed (e.g., `./scripts/idf.sh exec rm -rf "$bdir"` or running cleanup with fallback).
#### 4. `.gitignore` updates
* **Issue:** `.gitignore` currently only ignores `/build/`. If a lint run is interrupted (Ctrl+C) before cleanup, `build-lint-*/` and `.lint-compile-db.json` will show up as untracked files in `git status`.
* **Recommendation:** Add `/build-*/` and `/.lint-compile-db.json` to `.gitignore`.

Plantucha:
Fixed in `c1a26f9`, all four.

**1. CI ran a different gate from the one developers run.** This was the important one — the PR was inert exactly where it needed to work. `.github/workflows/lint.yml` now calls `./scripts/lint.sh` and defines no analysis of its own, so "green locally" and "green in CI" become the same claim rather than two claims that happened to agree.

Measured on this PR: the lint job goes 95s → **2m30s**, against the IDF build job's 3m21s. The jobs run concurrently, so the critical path is unchanged and the compile-database gate costs a contributor nothing in wall-clock time. `timeout-minutes` raised 15 → 30 to cover the image pull, and the path filter now includes `sdkconfig*.defaults` and `CMakeLists.txt`, since both change what gets built and therefore what gets analysed.

**2. The `/project` substitution.** Worse than "could mangle" — `str.replace()` destroyed the filename rather than relocating it:

```
/project/main/projects.c  ->  <root>/main<root>s.c
```

`rehost()` now requires the prefix *and* its trailing separator; `rehost_arg()` applies it per compiler argument, so `-I/project/main` is rewritten while `-I/opt/project/include` is left alone. Unit-tested against the shapes you named — a file called `projects.c`, one called `net_project.c`, a checkout under `~/projects`, and an unrelated `/opt/project` path. I also switched to `arguments` over `command` where the generator emits it, since it is already tokenised and cannot be mis-split on a path containing a space.

**3. Root-owned `build-lint-*`.** Reproduced here exactly as you described:

```
rm: cannot remove 'build-lint-t/CMakeFiles/git-data/head-ref': Permission denied
```

Cleanup now falls back to removing it from inside the container, and says so plainly if the directory survives, rather than leaving a silently dirty tree.

**4. `.gitignore`** — `/build/` → `/build*/` plus `/.lint-compile-db.json`. That also covers the `build-<board>/` case you raised on #251; I kept it in this branch only so the two PRs do not conflict on merge.

One thing I would rather state than have found: while making these changes I mis-anchored the heredoc carrying the embedded Python and left the old body in place. `bash -n` reported the script clean, because it does not parse heredoc bodies — the gate caught it by printing `NO COMPILE DATABASE` and falling back to the directory scan. The patch now compiles the embedded Python as part of applying it.

Rebased onto `3e8d371` and re-run: clean, all workflows pass.

ilyakruchinin:
Merged! 
Thank you, @Plantucha !

PR #250, opened 2026-09-09
test(mutate_host): print the denominator — most of the tree is invisible to it

Plantucha:
Follow-up to #232. The harness only mutates files a wired host test links, so everything else is not "clean" in its report — it is **unseen**, and a run ending in `0 unasserted survivors` says nothing whatever about it.

That was true and unstated, which is the part worth fixing. Measured: run against a branch adding 63 new `.c`/`.h` files under `main/`, the report came back byte-identical to the branch without them. Nothing was wrong; the harness simply could not see any of it, and said so nowhere.

Every run now prints its scope first:

```
SCOPE     3 of 80 C sources are linked by a wired host test; 77 are INVISIBLE to this harness
          not seen: components/therapy_alert/therapy_alert.c, … +72 more
          a clean report below says nothing about those files.
```

3 of 80 on `main` today.

A mutation score without its denominator is a comfort rather than a measurement, and this tool exists to argue the opposite — hiding its own denominator was the wrong shape for it.

Widening the `TESTS` map is a separate question and a bigger one: most of those files have no host test to link against. Reporting the gap is what stops the number being read as coverage it is not.

`--selftest` still discriminates. Sweep unchanged at ~35 s.

### DIFF
--- a/scripts/mutate_host.py
+++ b/scripts/mutate_host.py
@@ -221,6 +221,35 @@ def executed_lines(outdir: str) -> dict[str, set[int]] | None:
     return cov if cov else None
 
 
+def harness_blind_spot() -> tuple[int, int, list[str]]:
+    """(covered, total, a few uncovered names) for the C sources under main/ and components/.
+
+    ⚠️ THE HARNESS ONLY SEES WHAT A HOST TEST LINKS. TESTS above is a hand-written map, so a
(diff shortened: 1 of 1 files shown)


ilyakruchinin:
Merged! 
Thank you @Plantucha !

PR #251, opened 2026-09-09
ci(idf-build): build every board configuration, discovered rather than listed

Plantucha:
Follow-up to #218. The job compiles whichever board the tree's sdkconfig selects — which is one board. Every source behind `if(CONFIG_SOMNOTRACE_BOARD_X)` in `main/CMakeLists.txt` is then never compiled by CI at all: not built, not warned about, not linted, while the run goes green.

Not hypothetical. A 92,000-line branch proposed for this repo adds 32 files under `main/`, all behind such a guard. **A single-board run compiled zero of them and passed in 23 seconds.** Building the board they belong to surfaces, among other things, an out-of-bounds read of a 7-element array guarded by `< 8` on the OTA failure-reporting path.

## Discovered, not listed

The matrix comes from `sdkconfig*.defaults` rather than being written out. A hand-listed matrix is the same bug one level up: correct until someone adds a board and forgets this file, at which point CI is green on code nothing compiled. The `paths:` filters get the same treatment — `sdkconfig*.defaults`, so a new board's defaults file triggers the workflow that is supposed to cover it.

## Two things that are easy to get wrong

**`SDKCONFIG` is per leg.** `SDKCONFIG_DEFAULTS` is only read when no sdkconfig exists yet; once one is written it wins, and a second leg silently rebuilds the first leg's board while reporting the second leg's name. I hit exactly this while writing it — four legs each producing an identical translation-unit count, none of them the board asked for.

**Each leg verifies it built what it claims.** Every symbol in the leg's defaults must appear in the resolved sdkconfig or the leg fails. A leg that quietly falls back to the default board is worse than no leg, because it reports a board as covered while compiling something else.

`fail-fast` is off: which board broke is the information this job exists to give.

## Cost

This repo has one `sdkconfig*.defaults`, so today it is **a matrix of one and costs exactly what it costs now**. Verified on a tree with four: legs resolve to 55 / 65 / 54 / 66 translation units, the 7b leg sets `CONFIG_SOMNOTRACE_BOARD_WAVESHARE_7B=y` and produces a 2,908,912-byte image against the default leg's 2,463,408.

Not flashed — I have no ESP32-S3, so this verifies that each configuration **builds**, never that it runs.

### DIFF
--- a/.github/workflows/idf-build.yml
+++ b/.github/workflows/idf-build.yml
@@ -15,7 +15,7 @@ on:
       - "components/**"
       - "third_party/**"
       - "CMakeLists.txt"
-      - "sdkconfig.defaults"
+      - "sdkconfig*.defaults"
       - "partitions.csv"
       - "scripts/idf.sh"
       - "scripts/build-dist.sh"
@@ -27,7 +27,7 @@ on:
       - "components/**"
       - "third_party/**"
       - "CMakeLists.txt"
-      - "sdkconfig.defaults"
+      - "sdkconfig*.defaults"
       - "partitions.csv"
       - "scripts/idf.sh"
       - "scripts/build-dist.sh"
@@ -39,9 +39,50 @@ permissions:
   contents: read
 
 jobs:
+  discover:
+    # Emits one matrix leg per sdkconfig*.defaults the repo ships.
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.find.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v4
+      - id: find
+        run: |
+          set -o pipefail
+          python3 - <<'PY' | tee -a "$GITHUB_OUTPUT"
+          import glob, json, os
+          base = "sdkconfig.defaults"
+          legs = []
+          if os.path.exists(base):
+              legs.append({"name": "default", "defaults": base})
+          for f in sorted(glob.glob("sdkconfig.*.defaults")):
+              if f == base:
+                  continue
+              # sdkconfig.<name>.defaults -> <name>
+              name = f[len("sdkconfig."):-len(".defaults")]
+              legs.append({"name": name,
+                           "defaults": f"{base};{f}" if os.path.exists(base) else f})
+          if not legs:
+              # No defaults file at all: still build once, with whatever the tree implies,
+              # rather than silently building nothing.
+              legs = [{"name": "default", "defaults": ""}]
+          print("matrix=" + json.dumps(legs))
+          PY
+
+      - name: Say what will be built
+        run: echo '${{ steps.find.outputs.matrix }}'
+
   idf-build:
+    needs: discover
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    strategy:
+      # One board failing must not cancel the others: which boards break is the
+      # information this job exists to produce.
+      fail-fast: false
+      matrix:
+        config: ${{ fromJson(needs.discover.outputs.matrix) }}
+    name: "idf-build (${{ matrix.config.name }})"
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


ilyakruchinin:
Pre-checks failing

Plantucha:
Fixed in `515ff38`. The failing check was one I added, and it was wrong.

The leg-verification step required every symbol in a leg's defaults file to appear verbatim in the resolved sdkconfig, and that failed this repo's **default** board:

```
CONFIG_BT_NIMBLE_PINNED_TO_CORE=1 from sdkconfig.defaults did not take
CONFIG_MBEDTLS_DYNAMIC_BUFFER_PEER_CRED=y from sdkconfig.defaults did not take
```

Kconfig legitimately drops a default whose dependencies are unmet, so asserting IDF internals fails a correct configuration — which is the failure this job exists to prevent, one level up. My own check was the thing it was built to catch.

The assertion is now narrowed to `CONFIG_SOMNOTRACE_*`, which is the project's own namespace: a missing one means the leg fell back to a different board, rather than that Kconfig disagreed with a vendor default. The step reports how many symbols it verified so a leg checking zero is visible rather than silently vacuous.

Worth saying plainly: with one `sdkconfig*.defaults` in the repo this is a matrix of one and changes nothing about cost today. If you would rather hold it until there is actually a second board to build, that is a completely reasonable call and I will not be offended — the value only arrives with the second board.

ilyakruchinin:
Thanks @Plantucha! The dynamic discovery via `sdkconfig*.defaults` is well thought out—discovering board variants from the file tree rather than maintaining a static YAML matrix is the right pattern to prevent silent matrix drift. Scoping the validation assertion to `CONFIG_SOMNOTRACE_*` symbols is also the correct solution to avoid failing when Kconfig intentionally drops unmet IDF dependencies.
Here are our notes and recommendations:
#### 1. Timing vs. Repository State (Single-board today)
* **Rationale:** As you noted in your comment, SomnoTrace currently ships only `sdkconfig.defaults`. Today, this dynamic discovery produces a matrix of one. It introduces an extra workflow job (`discover`) with its own checkout and spin-up overhead (~5–10s) on every push/PR that touches firmware files.
* **Recommendation:** We agree with your assessment in the PR thread: holding this PR until the first multi-board PR (e.g., Waveshare 7B) lands makes complete sense. We can either:
  1. Hold and merge this PR simultaneously when the second board's `sdkconfig.*.defaults` is ready to be merged, or
  2. Merge this now so CI is pre-conditioned for multi-board contributions.
  If you have the second board branch ready, feel free to link it or submit the board defaults alongside this change.
#### 2. Verification on the default board
* **Observation:** `sdkconfig.defaults` currently has no `CONFIG_SOMNOTRACE_*` symbols (all project features are currently enabled via default macros or code guards). The verification step outputs:
      0 CONFIG_SOMNOTRACE_* symbol(s) verified in build-default/sdkconfig
  While harmless, the check is currently a no-op for the default board.
* **Recommendation:** Add a brief comment or ensure that when board definitions are introduced, they include at least one `CONFIG_SOMNOTRACE_BOARD_*` define so the validation step has symbols to assert.
#### 3. `.gitignore` updates
* **Issue:** Building into `build-${{ matrix.config.name }}` works cleanly on the CI runner, but if a developer tests locally with `-B build-<name>`, `.gitignore` currently only ignores `/build/`.
* **Recommendation:** Update `.gitignore` from `/build/` to `/build*/` so variant build trees do not appear in `git status`.

Plantucha:
Thanks — and on the timing question, option 2, though it is your call and I would be equally happy to have this sit.

We do not have a second board branch to link: no 7-inch hardware here, so anything I wrote for it would be untested by me. That is what makes holding a little awkward — "merge alongside the first multi-board PR" means this waits on someone else's work, and the failure it prevents is precisely the one that bites *when* that work arrives. Merging now costs the `discover` job's few seconds on firmware-touching pushes, and means the second board's contributor finds CI already able to build it rather than discovering the gap in review.

On the no-op verification: agreed, and it is at least a visible no-op — the step prints `0 CONFIG_SOMNOTRACE_* symbol(s) verified` rather than passing silently. I have deliberately not made a zero count fail, since that would red the default board today for no defect. When board symbols arrive the assertion starts doing work on its own, with no further change here.

`.gitignore` `/build/` → `/build*/` is in #249 rather than here, so the two do not conflict on merge. It covers `build-<board>/` as well as that PR's `build-lint-N/`.

Rebased onto `3e8d371`: clean, IDF build passes.

ilyakruchinin:
Merged! 
Thank you, @Plantucha !

PR #256, opened 2026-09-09
test(edf): the identification profile had no test, and the harness runs nowhere

Plantucha:
Three things, in the order they were found. The first is the reason for the other two.

## 1. `edf_generate_identification_files()` had no test at all

After #254 the harness could finally see `edf_header.c`, and its verdict was `VOID`: the canary lands on `edf_generate_identification_files()`, no test called it, so emptying the whole function body changed nothing and the harness refused a verdict on all 456 lines rather than report a clean sweep it had not earned. That is fail-closed working — but the file it was protecting us from is the one that writes the profile the AS11 reads back.

Five tests, written to pin what a mutant would change rather than to reach the function:

- the profile carries every key, including the two that are deliberately always blank
- a numeric `ProductCode` is coerced to a string, because the source writes it either way and the profile always carries a string
- **`ProfileVariantIdentifier` in the source becomes `ProfileVariationIdentifier` in the output.** It reads like a typo and is not. Pinned so that "fixing" it breaks a test instead of a device
- absent fields are present-and-empty or zero, never missing
- `Identification.crc` is exactly four bytes, little-endian, over the whole json
- a missing `identification.json` is refused **and writes nothing**

`edf_header.c` now produces a real verdict:

```
canary: edf_generate_identification_files() body emptied → killed
killed 15   stillborn 3   unasserted 5   unreached 2
```

**One note on the CRC test, because the first version of it was wrong in an instructive way.** It asserted that the four bytes on disk matched `edf_crc32_ieee()` over the json — which is self-referential: a mutation inside the CRC moves both sides of the comparison equally. The harness said so immediately, reporting the loop bounds at lines 50 and 52 and the shift at line 56 as surviving a test that appeared to cover them. The fix is an oracle the code cannot move: CRC-32/ISO-HDLC publishes a check value, `0xCBF43926` for `"123456789"`. With that added those three mutants die, and the count goes from 12 killed / 8 unasserted to 15 / 5.

The five that remain are in other functions and left for a separate pass: `edf_write_field`'s truncation at :71, the atomic-path guard at :84, `edf_read_bin_file`'s malloc at :125, and the two header offsets at :188 and :189.

## 2. The searching harness runs nowhere, so: weekly

`scripts/mutate_host.py` has been in the repo since #232 with nothing invoking it — the same gap #221 closed for `mutants.py` — and #254 only made it *able* to run on a runner. The two halves are not interchangeable, which is why this is a second workflow and not a step in `mutation.yml`:

| | | |
|---|---|---|
| `mutants.py` | **pins** | a fixed set that must all die, 15/15 today. Red means someone broke a working test. 30s, belongs on every push, and is where it already is |
| `mutate_host.py` | **searches** | generates mutants it has never seen. 40 unasserted survivors on a green tree, so as a gate it is permanently red |

A permanently red gate teaches people to ignore CI, which is the reasoning that already made lint's style tier advisory. So this reports and does not block. The one thing it **does** fail on is `VOID` — a surviving canary or a failing baseline means the harness could not detect a change it made on purpose, and every number in the log is then worthless. Reporting that as a quiet success is how a mutation score becomes a comfort.

**Why weekly and not on push, measured rather than assumed:** 391s for `--all` against 30s for `mutants.py`. The slowest per-push job here is the IDF build at about 200s and the jobs run concurrently, so a 391s job becomes the critical path and roughly doubles the wait on every pull request — to deliver a report nobody must act on before merging. Saturday 20:00 UTC puts the result in place for Sunday.

## 3. The survivor inventory is committed, not generated

A weekly report saying "40 survivors" is a number people read twice and then stop reading, and it carries no memory: nobody can tell whether this week's 40 are last week's 40.

`scripts/mutation_survivors.txt` is that memory, and it is **checked in**. The property worth having is not the trend — it is that a pull request weakening an assertion **adds a line here, in the diff, next to the change that caused it**, seen by a reviewer who never ran the sweep. A gitignored copy would exist only on the machine that made it, and two developers would hold different baselines without knowing. It is the argument `scripts/mutate_equivalent.txt` already won, applied to the open questions instead of the closed ones.

Verified deterministic before committing anything: two runs over the same tree give an identical classification, and the format carries no paths, timings or counts — the same `path:line:operator` convention as `mutate_equivalent.txt`, so there is one format rather than two.

**The refusal is the point.** A narrower environment finds fewer survivors: without cJSON the two EDF suites are skipped, `SCOPE` drops from 8 of 80 to 2 of 80, and every survivor in the five EDF modules is simply not looked for. Writing the inventory from that run would *delete* those entries, and the diff would read exactly like someone had fixed them. So `--write-inventory` requires `--all` and refuses where any suite is skipped or gcov is missing:

```
with cJSON      may write inventory: True
without cJSON   may write inventory: False — these suites could not be built:
                as11_events_test, edf_gen_test, edf_properties_test
```

Without the flag it compares instead, and when the environment is degraded it says the comparison is unreliable rather than presenting absences as fixes. Two runs on a clean tree: `INVENTORY 40 known, 0 new, 0 no longer found`.

## Also here: an apt fix, because CI went red today for none of these reasons

Every job that installs a package runs `sudo apt-get update`, and the `ubuntu-latest` image ships apt sources for Google Chrome and Microsoft that this project never installs from. Today one served a bad index and `apt-get update` exited 100:

```
E: Failed to fetch https://dl.google.com/linux/chrome-stable/.../Packages.gz  Hash Sum mismatch
##[error]Process completed with exit code 100
```

`host-tests`, `lint` and `mutation` went red together, at step 3 with step 4 skipped — a third-party repository reddening a firmware gate before any project code ran. It survived a re-run, so an outage rather than a flake.

The sources are **matched by content, not by filename**. My first attempt named `google-chrome.list`, removed nothing, and failed identically; the image actually uses `google-chrome.sources`, deb822 format. A workaround that silently matches nothing is the worst kind of wrong, so the loop echoes what it removes:

```
removing apt source this project does not use: /etc/apt/sources.list.d/microsoft-prod.list
removing apt source this project does not use: /etc/apt/sources.list.d/google-chrome.sources
```

Removed rather than swallowed with `|| true`, which would also hide a genuine failure to reach the Ubuntu archive — the one repository these jobs do depend on.

**Overlap worth knowing before you merge:** the identical fix is on #249 too, in `host-tests.yml` and `mutation.yml`, because that PR's own runs were failing the same way. Same text in both, and neither branch otherwise modifies those two files, so they merge in either order. `lint.yml` is hardened only on #249, where it is already being rewritten.

## What I could not test

`workflow_dispatch` requires the workflow file on the default branch, so the fork refused to dispatch the sweep:

```
HTTP 404: workflow mutation-sweep.yml not found on the default branch
```

The command it runs is verified (391s, real log, both guard conditions dry-run against a real and a simulated VOID log) and the YAML parses, but the job wiring itself is unexercised until it lands somewhere. If you would rather see it run before trusting the schedule, `workflow_dispatch` is on it, so one manual run after merge will tell you.

Host suite 5 run, 0 failed, 0 skipped; `edf_gen_test` 16 → 21 passing. Merges clean onto `3e8d371`.

### DIFF
--- a/.github/workflows/host-tests.yml
+++ b/.github/workflows/host-tests.yml
@@ -33,19 +33,26 @@ jobs:
 
       - name: Install libcjson
         run: |
-          # ⚠️ DROP THE RUNNER'S THIRD-PARTY APT SOURCES FIRST. The ubuntu-latest image ships
-          # lists for Google Chrome, Microsoft and Azure that this project never installs from.
-          # When one of them serves a bad index, `apt-get update` exits 100 and every job that
-          # installs anything goes red before reaching a line of project code. Measured
-          # 2026-09-09: `Hash Sum mismatch` on dl.google.com/linux/chrome-stable reddened
-          # host-tests, lint and mutation together, and survived a re-run — it is an outage,
-          # not a flake, and retrying does not clear it.
+          # ⚠️ DROP THE RUNNER'S UNRELATED APT SOURCES FIRST. The ubuntu-latest image ships
+          # apt sources for Google Chrome and Microsoft that this project never installs from.
+          # When one of them serves a bad index, `apt-get update` exits 100 and the job dies
+          # before reaching a line of project code. Measured 2026-09-09: `Hash Sum mismatch`
+          # on dl.google.com/linux/chrome-stable reddened host-tests, lint and mutation
+          # together, and survived a re-run — an outage, not a flake.
           #
-          # Removing the lists is the fix rather than `|| true`, which would hide a genuine
-          # failure to reach the Ubuntu archive — the one we do depend on.
-          sudo rm -f /etc/apt/sources.list.d/google-chrome.list \
-                     /etc/apt/sources.list.d/microsoft-prod.list \
-                     /etc/apt/sources.list.d/azure-cli.list
+          # MATCHED BY CONTENT, NOT BY FILENAME. The first version of this named
+          # /etc/apt/sources.list.d/google-chrome.list, removed nothing, and failed exactly as
+          # before: the image's file names differ between runner releases and between the
+          # .list and deb822 .sources formats. A guessed name fails silently, which is the
+          # worst way for a workaround to be wrong. Grepping for the host cannot miss.
+          #
+          # Removed rather than swallowed with `|| true`, which would also hide a genuine
+          # failure to reach the Ubuntu archive — the one repository these jobs depend on.
+          for f in $(grep -rlE 'dl\.google\.com|packages\.microsoft\.com' \
+                       /etc/apt/sources.list /etc/apt/sources.list.d/ 2>/dev/null || true); do
+              echo "removing apt source this project does not use: $f"
+              sudo rm -f "$f"
+          done
           sudo apt-get update -q
           sudo apt-get install -y --no-install-recommends libcjson-dev
 
--- a/.github/workflows/host-tests.yml
+++ b/.github/workflows/host-tests.yml
@@ -33,6 +33,19 @@ jobs:
 
       - name: Install libcjson
         run: |
+          # ⚠️ DROP THE RUNNER'S THIRD-PARTY APT SOURCES FIRST. The ubuntu-latest image ships
+          # lists for Google Chrome, Microsoft and Azure that this project never installs from.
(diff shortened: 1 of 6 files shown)


ilyakruchinin:
Merged! 
Thank you, @Plantucha !

Issue #228, opened 2026-09-08
A Small Finding about as11_ble reconnect retries


irfanguder:
When the CPAP device is not connected, during retries even if the logs say that it will retry in 60sec, actually it tries to reconnect every 90secs.
It is not a problem at all but in order to get rid of any inconsistencies I wanted to report.
Thanks.

I (10:46:15.660) as11_ble: reconnect: slow retry 633 in 60s...
W (10:47:45.660) as11_ble: reconnect: slow retry 633 failed: connection failed
I (10:47:45.660) as11_ble: reconnect: slow retry 634 in 60s...
W (10:49:15.660) as11_ble: reconnect: slow retry 634 failed: connection failed
I (10:49:15.660) as11_ble: reconnect: slow retry 635 in 60s...
W (10:50:45.659) as11_ble: reconnect: slow retry 635 failed: connection failed
I (10:50:45.660) as11_ble: reconnect: slow retry 636 in 60s...
W (10:52:15.660) as11_ble: reconnect: slow retry 636 failed: connection failed
I (10:52:15.660) as11_ble: reconnect: slow retry 637 in 60s...
W (10:53:45.660) as11_ble: reconnect: slow retry 637 failed: connection failed
I (10:53:45.660) as11_ble: reconnect: slow retry 638 in 60s...
W (10:55:15.660) as11_ble: reconnect: slow retry 638 failed: connection failed
I (10:55:15.660) as11_ble: reconnect: slow retry 639 in 60s...
W (10:56:45.660) as11_ble: reconnect: slow retry 639 failed: connection failed
I (10:56:45.660) as11_ble: reconnect: slow retry 640 in 60s...

ilyakruchinin:
Good catch, @irfanguder! You are 100% right about the 90-second cadence.

Here is why that happens:
1. The retry loop sleeps for **60 seconds** (`vTaskDelay`).
2. Then it calls `ble_gap_connect()`, which has a **30-second** connection establishment timeout when the CPAP is powered off.
3. Total elapsed time per cycle: `60s (wait) + 30s (BLE connect timeout) = 90s`.

Because the failure log only appears after the 30s connect timeout expires, each full retry attempt takes exactly 90 seconds. 

90s is actually a great interval when the CPAP is off because it conserves battery and radio airtime, but the log message `"reconnect: slow retry %d in 60s..."` is confusing because it sounds like the whole cycle repeats every 60s.

[...]

PR #264, opened 2026-09-12
test(as11_ble): extract the reconnect schedule and the AD salvage parser, and test both

Plantucha:
Two pieces of pure logic lifted out of `as11_ble.c` so they can be tested on the host. Neither changes behaviour; both are currently unreachable from any test because they live in a 3,000-line file with thirteen IDF includes.

## 1. The reconnect schedule (`as11_reconnect.c/.h`)

`reconnect_task` computes its backoff inline. The new module is the same arithmetic behind a name, with zero IDF includes:

```c
int as11_reconnect_delay_s(int attempt)
{
    if (attempt <= 1) return 0;
    if (attempt <= AS11_RECONNECT_FAST_ATTEMPTS) return attempt * 2;
    return AS11_RECONNECT_SLOW_DELAY_S;
}
```

Schedule: 1 → 0 s, 2 → 4 s, 3 → 6 s, 4+ → 60 s. `as11_ble.c` now reads both phases from it.

While writing the test I noticed the Phase 1 comment has drifted from the code:

```c
* Phase 1 — fast: 3 attempts with 2-4s backoff for transient disconnect
```

`attempt * 2` over attempts 2 and 3 gives **4 s and 6 s**, not 2–4 s. This is the block you clarified in `bca2236` for #228 — the Phase 2 half is now right, Phase 1 was not part of that change. The comment is corrected here; the timing is untouched.

`scripts/as11_reconnect_test.c` pins all four cases plus the boundaries.

## 2. The advertising-data salvage parser (`as11_adv.c/.h`)

`gap_event()` hand-walks AD structures when `ble_hs_adv_parse_fields()` rejects a packet — the AS11's `ADV_IND` carries a trailing type-`0x1b` field that is one byte short, so NimBLE discards the whole payload including the valid name. That walk is now a function.

The UUID cast stays at the call site: the 16-bit UUID list starts at an arbitrary offset with no alignment guarantee, so the cast belongs where the caller can see it, not hidden inside the parser.

`scripts/as11_adv_test.c` covers name/UUID extraction, zero-length AD, a length byte that overruns the buffer, and the off-by-one the parser's own comment warns about — the length byte counts the type byte, so a four-character name is `ad_len == 5`.

## Checks

Both suites are registered in `run_host_tests.sh` and in the mutation harness. Full host gate green, mutation clean.

Happy to split this into two if you would rather review them separately.

### DIFF
--- a/main/CMakeLists.txt
+++ b/main/CMakeLists.txt
@@ -1,4 +1,4 @@
-idf_component_register(SRCS "main.c" "bsp_power.c" "bsp_display.c" "net_provision.c" "as11_ble.c" "as11_reconnect.c"
+idf_component_register(SRCS "main.c" "bsp_power.c" "bsp_display.c" "net_provision.c" "as11_ble.c" "as11_reconnect.c" "as11_adv.c"
                            "sd_storage.c" "session_writer.c" "time_sync.c"
                            "post_therapy.c" "edf_gen.c" "edf_header.c" "edf_waveform.c" "edf_annotations.c" "edf_summary.c" "log_stream.c" "device_settings.c"
                            "psram_task.c" "bsp_audio.c" "bsp_i2c.c" "bsp_touch.c" "session_graph.c" "nvs_writer.c"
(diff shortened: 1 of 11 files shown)


ilyakruchinin:
Merged! 
Thank you, @Plantucha !

PR #265, opened 2026-09-12
test(host): sanitizers, a trailing-comment fix in the harness, and the eighteen edf_waveform guards

Plantucha:
**Read this first: the bottom five commits of this branch are #256.** GitHub cannot base a cross-fork PR on another PR's head, so they come along. Two ways to take it, whichever suits you:

- merge #256 first and this shrinks to the four commits below, or
- take this one and close #256 as included.

Nothing here is a second version of #256 — it is the same commits plus four more. I am posting it so the work is visible rather than sitting on my fork; there is no hurry on it.

## What is new beyond #256

**`e416fd1` — build the host suite with ASan and UBSan.** `run_host_tests.sh` probes whether the compiler supports `-fsanitize=address,undefined` and enables it when it does, with `SNT_NO_SANITIZE=1` to opt out. Two results, both measured on a 60-run sample:

- five more mutants die — memory-safety bugs that previously produced correct-looking output
- the sweep runs **3× faster**: 387 s → 126 s. A corrupting mutant now aborts in milliseconds instead of running to the 300 s timeout

**`fe8caba` — stop mutating operators inside trailing comments.** `mutate_host.py` was rewriting `+` and `<` that appear in `/* ... */` after code on the same line. Those mutants can never be killed because they change nothing, so they inflated the survivor count and cost a full build each.

**`4311440` — the `edf_waveform.c` event-file guards had no failing input.** Eighteen survivors lived in `edf_find_mask_on_time`, `edf_find_mask_off_time`, `edf_find_zle_edge_time` and `snt_available_samples`. Twelve are now killed by new cases; six are genuinely equivalent and recorded in `mutate_equivalent.txt` with the reason.

One of those tests found a real oracle defect on my side: the CRC test compared the file's bytes against `edf_crc32_ieee()`, so a mutation moved both sides of the comparison equally and three CRC-loop mutants survived a test that looked like it covered them. It now checks the published IEEE value `0xCBF43926`.

**`f1da00c` — `eq-zero-channels` was unreachable, not equivalent.** It was recorded as equivalent; it is actually dead code, which is a different thing and is classified differently by `mutants.py`'s own rule. Reclassified.

## Net effect on the survivor inventory

40 → 33 (sanitizers) → 15 (waveform tests). `scripts/mutation_survivors.txt` is tracked, so a new survivor lands in a diff rather than in a log nobody reads.

## Checks

Full host gate green, mutation gate green, exit codes checked rather than the last line of output.

### DIFF
--- a/.github/workflows/host-tests.yml
+++ b/.github/workflows/host-tests.yml
@@ -33,19 +33,26 @@ jobs:
 
       - name: Install libcjson
         run: |
-          # ⚠️ DROP THE RUNNER'S THIRD-PARTY APT SOURCES FIRST. The ubuntu-latest image ships
-          # lists for Google Chrome, Microsoft and Azure that this project never installs from.
-          # When one of them serves a bad index, `apt-get update` exits 100 and every job that
-          # installs anything goes red before reaching a line of project code. Measured
-          # 2026-09-09: `Hash Sum mismatch` on dl.google.com/linux/chrome-stable reddened
-          # host-tests, lint and mutation together, and survived a re-run — it is an outage,
-          # not a flake, and retrying does not clear it.
+          # ⚠️ DROP THE RUNNER'S UNRELATED APT SOURCES FIRST. The ubuntu-latest image ships
+          # apt sources for Google Chrome and Microsoft that this project never installs from.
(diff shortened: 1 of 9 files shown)


ilyakruchinin:
Merged!
Thank you, @Plantucha !

Issue #262, opened 2026-09-12
psram_task_create: the deferred-free contract in the header has no callers, so every self-deleting task leaks its stack and TCB


Plantucha:
`psram_task.h` already documents the hazard:

> For self-deleting tasks: the caller must keep the out_stack/out_tcb pointers and free them after the task has exited (e.g. on next invocation — "deferred free" pattern).

No caller does. I parsed all 36 `psram_task_create()` call sites in `main/` and `components/` with a balanced-paren walk rather than grep, so multi-line calls are included: **every one passes `NULL, NULL`** for `out_stack` and `out_tcb`. The contract has no users, and there is no other reaper — nothing in the tree frees a stack or TCB except on the creation-failure paths.

## Why nothing reclaims them

`psram_task_create()` allocates both buffers on the heap and hands them to the *static* creation API:

```c
StackType_t *stack = heap_caps_malloc(stack_size, MALLOC_CAP_SPIRAM);
StaticTask_t *tcb  = heap_caps_malloc(sizeof(StaticTask_t), MALLOC_CAP_INTERNAL);
...
xTaskCreateStaticPinnedToCore(task_func, name, stack_size, arg, priority, stack, tcb, core_id);
```

In the pinned `espressif/idf:v5.5.1`, that function stamps the task as statically allocated — `components/freertos/esp_additions/freertos_tasks_c_additions.h:360`:

```c
#if ( tskSTATIC_AND_DYNAMIC_ALLOCATION_POSSIBLE != 0 )
{
    pxNewTCB->ucStaticallyAllocated = tskSTATICALLY_ALLOCATED_STACK_AND_TCB;
}
#endif
```

and `prvDeleteTCB()` in `FreeRTOS-Kernel/tasks.c` then takes this branch:

```c
else
{
    /* Neither the stack nor the TCB were allocated dynamically, so
     * nothing needs to be freed. */
    configASSERT( pxTCB->ucStaticallyAllocated == tskSTATICALLY_ALLOCATED_STACK_AND_TCB );
    mtCOVERAGE_TEST_MARKER();
}
```

The flag's declaration says the same thing: *"Set to pdTRUE if the task is a statically allocated to ensure no attempt is made to free the memory."*

[...]

PR #271, opened 2026-09-17
fix(therapy_alert): the alert routine orphans 16 KB of PSRAM on every escalation

Plantucha:
`alert_routine_task()` ends in `vTaskDelete(NULL)`, and both of its buffers are caller-owned:

```c
/* components/therapy_alert/therapy_alert.c — start_alert_routine() */
579   StackType_t  *stack = heap_caps_malloc(16384, MALLOC_CAP_SPIRAM);
580   StaticTask_t *tcb   = heap_caps_malloc(sizeof(StaticTask_t), MALLOC_CAP_INTERNAL);
582   h = xTaskCreateStaticPinnedToCore(alert_routine_task, "alert_routine",
584                                     stack, tcb, 0);
```

`xTaskCreateStaticPinnedToCore()` stamps the task `tskSTATICALLY_ALLOCATED_STACK_AND_TCB`, so `prvDeleteTCB()` takes the branch that releases nothing — the same mechanism as #262. Nothing else in the file frees them, so each completed routine orphans **16 KB of PSRAM plus one `StaticTask_t`** of internal RAM.

`start_alert_routine()` runs once per escalation (`handle_therapy_stop()` while `ALERT_ARMED` and inside the window), so this recurs for the life of the boot rather than costing once. The PSRAM side is survivable; the `StaticTask_t` comes from the internal pool that `psram_task.h` exists to protect.

## Why `cebebe7c` does not already cover it

`cebebe7c` put the reaper inside `psram_task_create()`, which is the right place for the call sites that use it. This site is not one of them: `therapy_alert` is a component, `psram_task.h` lives in `main/`, and `components/therapy_alert/CMakeLists.txt` requires only `nvs_flash json esp_http_client esp_timer mbedtls`. A component reaching into `main` would invert the dependency, so the sweep has to live here.

For completeness, the other two hand-rolled sites are fine: `session_graph.c:1081` and `upload_sched.c:775` allocate the same way but their tasks never self-delete.

## The sweep

Four slots, no dynamic allocation, run from the owner task's existing 30 s tick and again before a new routine allocates.

Freeing waits for a **second** sweep to confirm `eDeleted`, for the same reason `psram_task.c` defers: `eTaskGetState()` reports `eDeleted` from the moment `vTaskDelete(NULL)` is called, which is *before* the idle task has finished with the TCB. Waiting one sweep keeps both buffers alive across that window. The state remains `eDeleted` afterwards as well — once the idle task unlinks the task, `xStateListItem` has no container, which `eTaskGetState()` also maps to `eDeleted` — so a finished routine is never missed and never freed early.

Overflow is deliberate rather than fatal. If all four slots are still occupied the routine **still starts** and logs a warning, leaving that one stack unreclaimed. Suppressing a therapy alert to protect 16 KB would be the wrong trade, and that path is no worse than current behaviour, where every routine leaks.

## Also fixed

The creation-failure path returned without freeing:

```c
if (stack && tcb) { h = xTaskCreateStaticPinnedToCore(...); }
if (!h) { ESP_LOGE(...); set_state(ALERT_DISARMED); return; }   /* stack/tcb leaked */
```

A successful stack allocation paired with a failed TCB allocation leaked even though no task was ever created. `session_graph.c:1086` and `upload_sched.c:781` already free on that path; this file did not.

## What I have not done

Run it on hardware. This is a source-level fix, verified with `./scripts/idf.sh build` against the pinned `espressif/idf:v5.5.1` — build clean, `somnotrace.bin` 43% of the app partition free.

The leak is measurable the way #262 described, if you want it confirmed on a device: note `ps_free` and `ps_lfb` from `/api/status`, then drive ten therapy-stop escalations inside the alert window. Expected before this change: `ps_free` down ~16 KB per escalation with `ps_lfb` falling at least as fast, neither recovering.


ilyakruchinin:
Merged!
Thank you, @Plantucha !

PR #272, opened 2026-09-17
test(as11_events): link the real event parser instead of a copy of it

Plantucha:
`scripts/as11_events_test.c` opened with this:

```c
/* Replicated event taxonomy from session_writer.c */
typedef enum { AS11_EV_NONE = 0, ... } as11_event_t;

/* Replicated parser from session_writer.c */
static as11_event_t check_event_notification(const cJSON *msg, const char **out_report)
```

It was a copy because the original is `static` inside `session_writer.c` — 2990 lines pulling in FreeRTOS, the SD stack and the uploader, so a host test could not link it.

I diffed the two: **they match today.** That is the whole problem. Nothing keeps them in sync, so the moment someone adds an event name, reorders a branch or fixes a bug in the real parser, `as11_events_test` keeps passing against the stale duplicate — and it reads as coverage while proving nothing about shipped code. Distinguishing Mask Fit and tube-drying Cooldown from real therapy is exactly the decision that must not drift quietly: a misclassification writes a session that never happened.

## The move

`main/as11_events.{c,h}`, following `as11_reconnect.c` and `as11_adv.c` from #264. `session_writer.c` includes the header; its call sites are untouched. The function keeps its name rather than gaining a prefix, so the diff stays reviewable — happy to rename if you would prefer `as11_events_*`.

Net: **127 lines of duplication deleted**, one translation unit with no ESP-IDF dependency beyond cJSON.

## Wiring it in was not just plumbing

`scripts/mutate_host.py` needed the source in its per-test map (the map is explicit by design — #202, #254). The first sweep of the newly visible file:

```
══ main/as11_events.c
   canary: check_event_notification() body emptied → killed
   killed 0   stillborn 1   unasserted 4   unreached 0   known-equivalent 0
     🔴 UNASSERTED  main/as11_events.c:35  logical:||→&&   if (!events || !cJSON_IsArray(events))
     🔴 UNASSERTED  main/as11_events.c:38  relational:<→<=  for (int i = 0; i < n; i++)
     🔴 UNASSERTED  main/as11_events.c:42  logical:||→&&   if (!event || !cJSON_IsString(event))
     🔴 UNASSERTED  main/as11_events.c:47  logical:&&→||   if (out_report && rt && cJSON_IsString(rt))
```

Every malformed-payload guard in the parser could be flipped with the suite still green. Three are real, and **two fault rather than merely misclassify**:

- **`:35`** — `"events"` present but an *object* rather than an array. `cJSON_GetArraySize()` does not type-check; it counts children, and `cJSON_GetArrayItem()` walks the same list. Without the `cJSON_IsArray()` guard, an object's members get classified as events.
- **`:42`** — `"event"` present but not a string. `valuestring` is NULL there, so dropping the `cJSON_IsString()` guard reaches `strcmp(NULL, ...)`.
- **`:47`** — a `TherapyStart` carrying a `reportTime`, with `out_report == NULL`. `session_writer.c` has call sites that pass NULL; the mutant writes through the null pointer. The existing tests passed NULL only for events that never reach this line.

Those are now Test 9. `:38` is **equivalent** — `cJSON_GetArrayItem(events, n)` returns NULL at or past the size and the body opens with `if (!ev) continue;`, so the extra iteration does nothing before the loop ends — and is recorded in `mutate_equivalent.txt` with the argument rather than silently suppressed.

After:

```
   canary: check_event_notification() body emptied → killed
   killed 3   stillborn 1   unasserted 0   unreached 0   known-equivalent 1
```

## Verification

- `main/as11_events.c`: killed 3, unasserted 0, known-equivalent 1, canary killed
- host suite: `8 run, 0 failed, 0 skipped`; `as11_events_test` 9/9
- firmware builds against the pinned `espressif/idf:v5.5.1` (`somnotrace.bin`, 43% of the app partition free)

Not run on hardware — the parser is pure, and nothing in this change reaches the radio.

One thing worth naming: the sweep still reports `11 of 86 C sources are linked by a wired host test; 75 are INVISIBLE to this harness`. This moves one file across that line. `session_writer.c` itself remains untestable for the same reason it always was, and the rest of it is untouched here.


ilyakruchinin:
Merged!
Thank you, @Plantucha !

PR #274, opened 2026-09-17
fix(edf_annotations): a duplicate event separated by another label survived dedup

Plantucha:
The dedup in `generate_annotation_edf()` compares each event only against the previous **kept** one:

```c
163  for (size_t i = 0; i < count; i++) {
164      if (dedup_count > 0 &&
165          ev_list[dedup_count - 1].onset_sec == ev_list[i].onset_sec &&
166          strcmp(ev_list[dedup_count - 1].label, ev_list[i].label) == 0) continue;
169      ev_list[dedup_count++] = ev_list[i];
```

That is correct only if identical `(onset, label)` pairs are adjacent by then. The sort above it ordered on `onset_sec` alone:

```c
151  for (size_t i = 0; i + 1 < count; i++)
152      for (size_t j = i + 1; j < count; j++)
153          if (ev_list[j].onset_sec < ev_list[i].onset_sec) { /* swap */ }
```

It is a selection sort, not a stable one, so events sharing a second have no defined relative order. `[A, B, A]` at one onset can come out in that order, each element is compared only against the previous survivor, and **both copies of A reach EVE.edf/CSL.edf** — the annotation is written twice.

## The fix

Break the onset tie on `label` — the same key the dedup compares on — so identical pairs are guaranteed adjacent and the existing pass becomes correct. One added condition.

## Making it testable

The sort and dedup move into `sort_and_dedup_events()`. No new module is needed: `scripts/edf_gen_test.c:53` already does `#include "edf_annotations.c"`, so a static helper is directly reachable from the suite.

Two tests:
- **the interleaved repeat** — `[Obstructive Apnea, CSR Start, Obstructive Apnea]` at onset 100 must come out as two events, not three
- **ordering with distinct events** — five events across three onsets, one exact repeat; asserts only the repeat is dropped, that label breaks the tie, that onsets ascend, and that an empty list is handled

`mutate_host.py main/edf_annotations.c` now reports **9 killed** where this path previously killed none:

```
canary: generate_annotation_edf() body emptied → killed
killed 9   stillborn 0   unasserted 4   unreached 12   known-equivalent 0
```

## Survivor inventory

`scripts/mutation_survivors.txt` follows the two existing survivors to their new line numbers (`:68→:105`, `:83→:120`), tracks the sort-loop bound into the helper (`:151→:65`), and records one new survivor with its reasoning:

```
main/edf_annotations.c:69:relational:<→<=   # before = strcmp(ev[j].label, ev[i].label) < 0;
```

Swapping two events that compare equal reorders entries identical in **both** sorted keys, differing at most in `dur_sec`, and the dedup keeps one of them either way — so which survives is *unspecified* rather than wrong. Asserting it would pin an arbitrary choice. Making it defined needs a decision about whether a repeated event's longer or shorter duration should win, and that seemed yours rather than this PR's — happy to add it if you have a preference.

## Verification

- `edf_gen_test`: 40 passed (was 38), 0 failed
- host suite: `8 run, 0 failed, 0 skipped`
- firmware builds against the pinned `espressif/idf:v5.5.1`

## Not in scope

`scripts/as11_events_test.c` Test 8 carries its own inline copy of this dedup loop, so it asserts against a replica rather than the real one — the same pattern #272 removed for the event parser. It does not cover the interleaved case either way. Worth a follow-up; left alone here to keep this diff to the defect.

Not run on hardware: the changed code is pure list manipulation and nothing in it reaches the radio or the SD card.


ilyakruchinin:
Merged!
Thank you, @Plantucha !

Issue #273, opened 2026-09-17
StreamData gaps of 10-120 s are neither padded nor split, so the waveform EDF timeline silently compresses and annotations slide off it


Plantucha:
A StreamData gap of **10.1 s – 120 s** is neither compensated nor split. The samples are simply absent from a stream the EDF declares uniform, so everything after the gap is written as if it happened earlier than it did.

## The window

`session_writer.c` classifies gaps in three bands:

```c
2070   if (gap > 280) {
2071       int missing = (int)((gap - 100) / 200);
2072       if (missing > 0 && missing < 50) {          /* < 10.1 s — hold last value */
...
2118       } else if (missing >= 50) {                  /* >= 10.1 s — not compensated */
2119           /* Uncompensated discontinuity — record it honestly. */
2122           s->gap_long_events++;
2123           s->gap_long_ms_total += (uint64_t)gap;
2126           ESP_LOGW(TAG, "long StreamData gap: %lld ms (%d notifications) — not compensated");
2128           if (gap >= SW_SPLIT_GAP_MS) want_split = true;   /* >= 120 s — split */
```

Below 10.1 s the previous value is held. At or above 120 s the session splits, and the reasoning given there is exactly right:

> `2139` *Split rather than emit a continuous-looking timeline across a multi-minute hole. .snt assumes uniform sampling, so a gap this large cannot be represented honestly inside one session.*

Between those two bounds, neither happens.

## Why that is not neutral

The waveform EDFs are written as **continuous**:

```c
main/edf_waveform.c:304    0, record_dur, "EDF", sig, n_signals)
main/edf_waveform.c:351    "EDF", sig, n_signals);
```

Only EVE/CSL use `EDF+D` (`edf_annotations.c:197`). A plain `EDF` reserved field asserts uniform, contiguous sampling — sample *N* is at `start + N/rate`, and there is no per-record timestamp to say otherwise.

And the record count is derived from the sample count alone, with no time basis:

```c
main/edf_waveform.c:287    int total_records = (int)(total_samples / spr[0]);
```

So a session that lost 45 s of notifications produces a file that claims to be 45 s shorter than the session was, and every sample after the gap is attributed to a wall-clock time 45 s early. Gaps accumulate within a session.

[...]

PR #279, opened 2026-09-19
fix(session_writer): the split threshold and the pad ceiling disagreed by 100 ms

Plantucha:
## What

`SW_SPLIT_GAP_MS` is `10000`, but the branch that consults it is only reachable once `missing >= 50`, and `missing = (gap - 100) / 200` does not reach 50 until **gap is 10100 ms**. The pad ceiling and the split threshold therefore disagree by 100 ms.

## Why it matters

**1. A residual dead zone, in the range the 10 s threshold was introduced to close.**

A gap of 10000–10099 ms took the *pad* branch and was filled with 49 sample-and-hold notifications — 245 flow/pressure samples, ~9.8 s of synthetic 25 Hz waveform. That is precisely the fabricated physiological data the comment directly above it says must never be produced, for the reason given there: it manufactures artificial apneas in OSCAR / SleepHQ scoring.

**2. The constant governed nothing at its shipped value.**

```c
} else if (missing >= 50) {          // implies gap >= 10100
    ...
    if (gap >= SW_SPLIT_GAP_MS) want_split = true;   // 10000 — never false here
```

Changing `SW_SPLIT_GAP_MS` anywhere in 1..10100 produced no behavioural difference at all. Someone tuning it from 10000 to, say, 10050 would observe nothing and reasonably conclude the mechanism was broken.

## The change

Both branches now gate on the same quantity — `gap` against `SW_SPLIT_GAP_MS` — so they are exhaustive, cannot overlap, and cannot leave a hole. `missing` keeps its remaining job: sizing the padding loops and the log line. The vacuous inner test is dropped.

## Verification

Exhaustive sweep, 0..30000 ms at 1 ms resolution, old decision vs new:

```
CHANGED range: 10000..10099 ms  (100 values)
old -> new   : PAD -> SPLIT

  gap=9999    missing=49   old=PAD     new=PAD
  gap=10000   missing=49   old=PAD     new=SPLIT
  gap=10099   missing=49   old=PAD     new=SPLIT
  gap=10100   missing=50   old=SPLIT   new=SPLIT
  gap=120000  missing=599  old=SPLIT   new=SPLIT
```

Every other input is unchanged. "Record the long gap but do not split" was unreachable before this change and remains unreachable after it — so no path loses its `gap_long_*` accounting.

**No functional change for gaps below 10 s or at/above 10.1 s.**

## Comments

The constant read `/* 10 s: >= 50 missing notifications */`. Those are 10000 ms and 10100 ms respectively, so it asserted an equivalence that does not hold; it and the block comment now refer to the constant itself rather than restating a number that can drift from it.

## Not included

I did not add a host test. `session_writer.c` pulls in 9 ESP-IDF headers and is coupled to FreeRTOS mutexes and SD storage, so covering this under `scripts/` would mean extracting the gap decision into a pure function — a refactor rather than a fix, and your call rather than mine. If you'd like that as a follow-up, `sw_gap_action(int64_t gap) -> {NONE, PAD, SPLIT}` would be trivially testable against the sweep above and would keep the two thresholds provably in step.

Follow-up to #273.


Plantucha:
CI is green on all six checks.

Offering the follow-up I mentioned, concretely, so it's easy to decline.

**The problem it would solve.** This bug existed because the pad ceiling and the split threshold were two numbers that *nearly* agreed, expressed in different units — a notification count and a millisecond value — with no test holding them in step. Fixing the conditions removes today's 100 ms gap, but nothing stops them drifting apart again the next time either is tuned.

**Shape.** A tiny header with no ESP-IDF dependencies:

```c
/* main/session_gap.h */
typedef enum { SW_GAP_NONE, SW_GAP_PAD, SW_GAP_SPLIT } sw_gap_action_t;

/* Pure: what to do about a StreamData discontinuity of `gap` ms.
 * `missing_out` (optional) receives the notification count used to size padding. */
static inline sw_gap_action_t sw_gap_action(int64_t gap, int *missing_out)
{
    int missing = (gap > 280) ? (int)((gap - 100) / 200) : 0;
    if (missing_out) *missing_out = missing;
    if (gap >= SW_SPLIT_GAP_MS) return SW_GAP_SPLIT;
    if (missing > 0)            return SW_GAP_PAD;
    return SW_GAP_NONE;
}
```

`session_writer.c` includes it and switches on the result; the call site keeps its existing bodies. The header is the only thing the test needs, so `scripts/session_gap_test.c` links **the real decision rather than a copy of it** — the same reason you took #272.

**What the test would assert**, beyond a boundary table:

- `SW_GAP_PAD` and `SW_GAP_SPLIT` are exhaustive and disjoint across 0..600000 ms — no input can be both, and none can fall through. That is the invariant this bug violated, and it holds for *any* value of `SW_SPLIT_GAP_MS`, so it keeps working if you retune it.
- Padding is never emitted at or above `SW_SPLIT_GAP_MS` — the clinical rule from your #273 analysis, stated as an assertion instead of a comment.
- A sweep pinning the exact transition point, so a future change to either threshold fails loudly rather than silently moving the boundary 100 ms.

**Cost:** one small header, one mechanical edit at the call site, one test file, one line in `run_host_tests.sh`. No behaviour change on top of this PR.

**Happy for the answer to be no** — it is a refactor of code that now works, and you may prefer the call site to stay literal. If you'd rather have it, say so and I'll open it as a separate PR against main after this one lands, not folded into this one.

ilyakruchinin:
Merged!
Thank you, @Plantucha !

PR #280, opened 2026-09-20
fix(oximetry): Format-A cadence came from a constant, not from the file

Plantucha:
## What this changes

A Format-A recording states its own seconds-per-sample interval in its trailer. The converter ignored it and wrote 1 Hz as a literal in three places:

| Site | Was |
|---|---|
| `write_snt3_format_a` | `hdr.period_num_us = 1000000;` |
| `build_generation_manifest` | `cJSON_AddNumberToObject(track, "period_num_us", 1000000);` |
| `oximetry_canonical_convert_format_a` | `end_ms = start_utc_ms + (count - 1) * 1000;` |

The interval is a writable device setting, so a ring configured to anything other than 1 s produced sample timestamps that drift further apart the longer the recording runs, with 1 Hz asserted as fact in the manifest.

**The VLD3 path in this same tree already does the right thing.** `ox_vld3_parse_header` derives its period from the file and refuses it when the arithmetic does not land on a plausible cadence. Format-A asserted a constant instead. Same repository, same data, two standards, and this only makes the second match the first.

## Honest framing

**This is hardening, not a field fix.** Every stored file observed so far carries `interval = 1`, so nothing out there is currently wrong. What changes is that the code stops depending on that staying true. If you would rather carry it as a known assumption than as code, closing this is a perfectly reasonable call and I will not re-raise it.

## How it is built

New standalone unit, deliberately shaped like `oximetry_vld3.c` so it is host-testable the same way:

```
main/oxyii_trailer.{c,h}   48-byte trailer: finalisation magic at +4,
                           u32 sample count at +12, u8 interval at +16.
                           Depends only on <stdint.h> and <string.h>.
```

**The interval is validated, never trusted.** `0`, or anything above 60 s, leaves `period_us` at 1 Hz and clears `interval_valid`, so a damaged trailer degrades to the previous behaviour rather than fabricating a cadence. The conversion logs which of those happened instead of silently choosing.

**The trailer also states the recording's length, which the converter had no way to check.** It now compares that against the body-derived count and logs a disagreement. The body stays authoritative, because it reflects what was actually read — the point is only that a truncation the file size cannot reveal no longer passes in silence. I deliberately did not make it fatal: refusing a night because two numbers disagree is a worse failure than recording that they did.

## The test

`scripts/oxyii_trailer_test.c`, wired into `run_host_tests.sh` next to the VLD3 decoder test.

**The count is read as u32, and the test pins that specifically.** A 16-bit read truncates any session longer than 65,535 samples with nothing downstream able to notice — 70,000 comes back as 4,464. I verified the test has teeth rather than assuming it: reverting the upper two bytes of the count reds three assertions.

```
$ gcc ... scripts/oxyii_trailer_test.c mutant-with-16-bit-count.c -o t && ./t
  FAIL scripts/oxyii_trailer_test.c:99: count above 65535 must not truncate
  FAIL scripts/oxyii_trailer_test.c:100: a 16-bit read would have wrapped
  FAIL scripts/oxyii_trailer_test.c:104: count must be little-endian across 4 bytes
oxyii_trailer tests FAILED (3)
```

It uses an explicit failure counter rather than `assert()`, so it cannot be built into a run that reports success having checked nothing.

Local run of your harness:

```
sanitizers: address,undefined
### oxyii_trailer_test: PASS  (oxyii_trailer tests passed)
host tests: 7 run, 0 failed, 2 skipped
```

The two skips are `edf_gen_test` and `edf_properties_test`, for a missing local cJSON, not for anything in this change.

## What I could not verify

**I cannot build the firmware here** — no ESP-IDF toolchain on this machine, and `oximetry_canonical.c` pulls in FreeRTOS headers. The `idf-build` job is the authority on whether the signature changes compile on-target. The edits are mechanical (two signatures gained a `uint32_t period_us`, two call sites pass it), but I would rather say so than imply I proved it.

## Offered, easy to decline

If the cross-check proves useful, the natural follow-up is to surface it in the manifest's `integrity` block alongside `declared_size_matches`, so a consumer can see that the file's own count agreed rather than having to trust that nobody logged a warning. I have not done that here because it changes the manifest schema, which is your call and not mine.


ilyakruchinin:
Merged!
Thank you, @Plantucha !

PR #281, opened 2026-09-20
test(edf_gen): the MaskOn/MaskOff drift correction had no test that could tell the direction

Plantucha:
## The gap

`_ZLE` is the primary gating signal, and its AS11→NTP clock correction has a test. **The MaskOn fallback and the MaskOff end carry the same correction and had none that could tell one direction from the other.** Reversing either sign moves a session boundary by twice the drift, and the entire suite still passed.

Both are in the committed inventory:

```
main/edf_gen.c:270:arithmetic:+→-   int64_t maskon_ntp = maskon_as11 + clock_drift_ms;
main/edf_gen.c:312:arithmetic:+→-   end_ntp = maskoff_as11 + clock_drift_ms;
```

**The fallback deserves the coverage more than the path in front of it.** MaskOn is what runs when `_ZLE` is absent — older firmware, a failed subscription, a night the AS11 never gated. The untested path is the degraded one.

## The tests

Both observe through artefacts a user would actually see, not internals.

**`the MaskOn fallback applies the AS11->NTP clock drift`** — MaskOn at AS11 +7 s with a +3 s drift gates at NTP +10 s. The wrong instant is named and asserted **absent**, not just the right one asserted present: a test that only checks the expected file exists still passes if the exporter writes both.

**`MaskOff applies the AS11->NTP clock drift when ending the export`** — reversing the MaskOff sign leaves the start untouched, so no filename reveals it. This one reads the record count instead. The 30 s drift is chosen so each error has a distinct signature rather than because it is typical: reverse MaskOff and the export ends a record early; reverse MaskOn and the gate lands on the session start, is rejected as out of range, and the export is never gated at all.

Neither pins an expected sample value. The ungated export is the oracle, as in the `_ZLE` test they sit beside.

## Verified by mutation, not assumed

Each mutant applied to an otherwise clean tree:

```
baseline                          42 passed, 0 failed          exit 0

:270  maskon_as11 - clock_drift_ms
  FAIL :988 no BRP.edf at the drift-corrected MaskOn time (22:00:10)
  FAIL :989 an export exists at 22:00:04 - the drift was SUBTRACTED
                                  40 passed, 2 failed          exit 1

:312  maskoff_as11 - clock_drift_ms
  FAIL :1050 records: got 1 want 2 - one means MaskOff ended the export 2x the drift early
                                  41 passed, 1 failed          exit 1

restored                          42 passed, 0 failed          exit 0
```

Full suite with every backend available: `host tests: 8 run, 0 failed, 0 skipped`.

## A note on the inventory

**`scripts/mutation_survivors.txt` is hand-edited to drop those two lines rather than regenerated**, and I think that is the right call here.

I ran `mutate_host.py --all --write-inventory` first. It produces the correct count, but it also reorders unrelated entries and **deletes the five-line rationale above the `edf_annotations` dedup survivors** — the one explaining that the ordering is unspecified rather than wrong, and that fixing it needs a decision about whether a repeated event's longer or shorter duration should win. That is a human judgement the generator cannot reproduce, and a test PR should not quietly take it out.

The hand-edit agrees with what the harness reports independently: **16 → 14**.

If you would rather have the regenerated file and re-add the note yourself, say so and I will swap it.


ilyakruchinin:
Merged!
Thank you, @Plantucha !

PR #282, opened 2026-09-20
test(edf_gen): the export's three entry guards had no test that could tell them apart

Plantucha:
> **Stacked on #281.** That one covers the gating path; this one covers the entry path. The diff below includes #281's two commits, so review it first. Happy to rebase once it lands.

## What survived, and why

Three guards on the way into the export, each surviving the sweep for the same reason: **no test ever supplied the one input that separates the operator from its mutant.**

```
main/edf_gen.c:111:logical:||→&&      if (!session_dir || !session_id || !out_root)
main/edf_gen.c:126:relational:<→<=    if (start_epoch_ms < 946684800000LL)
main/edf_gen.c:161:arithmetic:-→+     as11_time_noon_day(start_epoch_ms - clock_drift_ms, ...)
```

**`a single NULL argument is refused`** — every existing caller passes all three pointers, so the guard was never exercised. Under `&&` a lone NULL walks into the body instead of being refused.

**`the year-2000 guard excludes only what precedes it`** — the bound is exclusive deliberately: the first instant of 2000 is a legitimate start, not the crash-recovery artefact the guard exists to catch. Only a start exactly on the bound separates `<` from `<=`. Asserted as *not refused for this reason* rather than `ESP_OK`, since whether a Y2K session exports cleanly is a different question from whether the guard lets it through.

**`the DATALOG day folder is computed in the AS11 clock`** — the one I think is worth having on its own merits. The folder is a noon-to-noon day in the *device's* clock, so the drift is **subtracted** here, the opposite direction from every other use of it in this file. That asymmetry is correct and was entirely unguarded. The 12 h drift is chosen for observability rather than realism, and the comment says so: only a drift large enough to carry the session across a noon boundary can reveal which direction was taken.

## Verified by mutation

```
baseline                                    45 passed, 0 failed    exit 0

:111  !session_dir && !session_id
  FAIL a NULL session_dir must be refused on its own
  FAIL a NULL session_id must be refused on its own              exit 1

:126  start_epoch_ms <= 946684800000LL
  FAIL midnight on 2000-01-01 is a valid start and must not hit the guard
                                                                  exit 1

:161  start_epoch_ms + clock_drift_ms
  FAIL no DATALOG/20260228 — the drift was not subtracted
  FAIL DATALOG/20260301 exists — the drift was ADDED              exit 1

restored                                    45 passed, 0 failed    exit 0
```

`host tests: 8 run, 0 failed, 0 skipped`.

## The remaining survivors, and why I stopped here

**16 on main → 11 with this.** I looked at all ten in `edf_gen.c` and deliberately left seven alone, because I think several are equivalent rather than gaps:

- **`:258`, `:316`** guard log lines and no-op branches — the two forms differ only in *which* message is emitted, never in what is exported.
- **`:269`, `:311`, `:309`** turn on `> 0` versus `>= 0` for values whose absent-sentinel is **−1**, not 0 (`edf_find_mask_on_time` and friends all `return -1`). Both forms take the same branch, so no input separates them.
- **`:245`, `:271`** are boundary-exact on the gating comparisons and probably *are* killable with an event landing precisely on `end_epoch_ms`, but the resulting zero-length export is a messier thing to assert than it is worth in the same change.

Your `mutate_equivalent.txt` asks for the argument to be written down rather than the entry just added, and I have not done the work to make those claims rigorously — "I could not construct an input" is not the same statement, as that file itself says. Happy to take that on separately if you want them settled either way.


ilyakruchinin:
Merged!
Thank you, @Plantucha !

